PR #47, opened 2018-11-30
Fix links

evqna:


evqna:
Looks like this will be fixed by #49

PR #48, opened 2018-11-30
Integrate gosec into the app

MVrachev:
Because we depend on the GOPATH enviormental variable and it's conditions I created a /go/src folders in cache folder and because I wanted to use the functions in cache.js I had to change almost all functions in cache.js to behave differently based on which app is calling them (Gosec or Bandit).

Also I added the lines in index.js in which we call Gosec, convert the output from Gosec and merge the outputs from Bandit and Gosec into one. Also I changed a few things about how we save the files initially.

I added the cwd (current working directory) flag when spawning gosec in gosec/gosec.js. This helped me to fix bugs when calling gosec and because of this I had to change where is the file I am removing in the test/gosec.spawn.test.js file.

And in the end I changed the format of the bandit summary so it will be easier to merge the outputs of Bandit and Gosec.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "standard": "^12.0.1"
   },
   "engines": {
-    "node": ">= 11.3.0"
+    "node": ">= 8.10"
   },
   "jest": {
     "testEnvironment": "node",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "standard": "^12.0.1"
   },
   "engines": {
-    "node": ">= 8.3"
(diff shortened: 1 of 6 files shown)


evqna:
This should be rebased on top of master to check if it passes the CI

package.json
```diff
@@ -41,7 +41,7 @@
     "standard": "^12.0.1"
   },
   "engines": {
-    "node": ">= 8.3"
+    "node": ">= 11.3.0"
```
evqna:
```suggestion
    "node": ">= 8.10"
```

PR #49, opened 2018-11-30
Changes in .travis.yml

MVrachev:
We need to change the Travis CI config file - .travis.yml in order to run tests on travis correctly.
This update correctly configures node versions for travis and installs bandit and gosec.
We use Node versions 8, 10 and 11 (all suported versions according to eslint) because that way we are sure our app works on all supported versions.

Antoine Salon's (@evqna) pull request https://github.com/vmware/precaution/pull/50 helped me a lot to finish this pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,24 @@
 # Copyright 2018 VMware, Inc.
 # SPDX-License-Identifier: BSD-2-Clause
 
-sudo: false
+dist: "trusty"
+
+install:
+  - yarn install
+  - pip3 install --user -r requirements.txt
+  - curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $GOPATH/bin 1.2.0
+  - export GOPATH=$GOPATH:/home/travis/build/vmware/precaution/test/fixtures/go
 language: node_js
 node_js:
-  - "8.3"
+  - "8"
+  - "10"
+  - "11"
+
+addons:
+  apt:
+    packages:
+      - "python3"
+      - "python3-pip"
+
 notifications:
   disabled: true


evqna:
Looks like I approved too quickly, it seems this configuration creates three independent test runs (one that installs the python dependencies, another with go and another with node) instead of installing all three and then running the tests.

.travis.yml
```diff
@@ -1,9 +1,22 @@
...
 
-sudo: false
-language: node_js
-node_js:
-  - "8.3"
+
 notifications:
   disabled: true
+matrix:
+  include:
```
ericwb: By using a matrix, you are effective creating 3 jobs.  One for go, one for python, and one for node.js.

But if you want to do install work as part of the node.js job, you use the `before_install` or `install` keyword for that job. I see any reason to use a matrix yet.

https://docs.travis-ci.com/user/job-lifecycle/
https://docs.travis-ci.com/user/build-matrix/

.travis.yml
```diff
@@ -1,9 +1,22 @@
...
+    - export PATH=$PATH:$HOME/gopath/bin
+  - language: python
+    python: 3.6.6
+    install:
+    -  pip install -r requirements.txt
+    script: exec ls -alR ${GOPATH}
+  - language: node_js
+    node_js: "11.3.0"
+notifications:
+  disabled: true
```
ericwb: Add proper new line

.travis.yml
```diff
@@ -1,9 +1,22 @@
...
+
+install:
+  - yarn install
+  - pip3 install --user -r requirements.txt
+  - curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $GOPATH/bin 1.2.0
+  - export GOPATH=$GOPATH:/home/travis/build/vmware/precaution/test/fixtures/go
 language: node_js
 node_js:
-  - "8.3"
+  - "11"
```
evqna: Can you put all 3 currently supported Node releases (8, 10 and 11)?
Travis will use the latest release for each so it should pass (previously it failed because ESlint requires 8.10 or higher).

evqna: Can you put all 3 currently supported Node releases (8, 10 and 11)?
Travis will use the latest release for each so it should pass (previously it failed because ESlint requires 8.10 or higher).

PR #50, opened 2018-12-01
Travis updates

evqna:
This update correctly configures node versions for travis and installs bandit and gosec.

All bandit tests pass but gosec ones are still failing, probably due to some gopath issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,22 @@
 # Copyright 2018 VMware, Inc.
 # SPDX-License-Identifier: BSD-2-Clause
 
-sudo: false
+dist: "trusty"
+install:
+  - yarn install
+  - pip3 install --user -r requirements.txt
+
 language: node_js
 node_js:
   - "8"
   - "10"
   - "11"
+
+addons:
+  apt:
+    packages:
+      - "python3"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,9 +1,24 @@
...
+
+install:
+  - yarn install
+  - pip3 install --user -r requirements.txt
+  - curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $GOPATH/bin 1.2.0
+
 language: node_js
 node_js:
-  - "8.3"
+  - "8"
```
ericwb: Martin's PR in https://github.com/vmware/precaution/pull/48/files is setting the minimum node version to 11.3.  So do we still need testing of 8 and 10?

evqna: We don't "need" it but since all of them are currently supported releases I think we should try to remain compatible. Especially since we are not using any features from node 11.x as far as I know.

evqna: Also the CI used to fail on 8.3 because ESLint requires 8.10 at minimum, just setting the major release number defaults to the latest release on that branch.

evqna:
Solved by #49

PR #54, opened 2018-12-05
Use different cache subfolders per repo

evqna:
Since pull request numbers are only unique per repository there might be name conflicts when multiple requests are processed simultaneously.


MVrachev:
It looks good to me.

PR #56, opened 2018-12-06
Integrate gosec into the app

MVrachev:
Because we depend on the GOPATH enviormental variable and it's conditions I created a /go/src folders in cache folder and because I wanted to use the functions in cache.js I had to change almost all functions in cache.js to behave differently based on which app is calling them (Gosec or Bandit).

Also I added the lines in index.js in which we call Gosec, convert the output from Gosec and merge the outputs from Bandit and Gosec into one. Also I changed a few things about how we save the files initially.

I added the cwd (current working directory) flag when spawning gosec in gosec/gosec.js. This helped me to fix bugs when calling gosec and because of this I had to change where is the file I am removing in the test/gosec.spawn.test.js file.

And in the end I changed the format of the bandit summary so it will be easier to merge the outputs of Bandit and Gosec.

### DIFF
--- a/cache.js
+++ b/cache.js
@@ -9,9 +9,16 @@ const path = require('path')
  * @param {number} repoID repository identifier
  * @param {number} prID pull request identifier
  * @param {string} branchTag branch name
+ * @param {string} app Optional: an app that calls the function. (default bandit)
+ * It's needed because gosec uses GOPATH.
  */
-function getBranchPath (repoID, prID, branchTag) {
-  return path.join('cache', repoID.toString(), prID.toString(), branchTag)
+function getBranchPath (repoID, prID, branchTag, app) {
+  app = app || 'bandit'
+  if (app === 'bandit') {
+    return path.join('cache', repoID.toString(), prID.toString(), branchTag)
+  } else if (app === 'gosec') {
+    return path.join('cache/go/src', repoID.toString(), prID.toString())
+  }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


cache.js
```diff
@@ -9,9 +9,16 @@ const path = require('path')
  * @param {number} repoID repository identifier
  * @param {number} prID pull request identifier
  * @param {string} branchTag branch name
+ * @param {string} app Optional: an app that calls the function. (default bandit)
+ * It's needed because gosec uses GOPATH.
  */
-function getBranchPath (repoID, prID, branchTag) {
-  return path.join('cache', repoID.toString(), prID.toString(), branchTag)
+function getBranchPath (repoID, prID, branchTag, app) {
+  app = app || 'bandit'
```
evqna: Wouldn't it be easier to set 'cache' as the GOPATH, then create a 'src' subfolder inside? That way we don't have to deal with different access paths for each linter

cache.js
```diff
@@ -31,9 +38,16 @@ function branchPathExists (repoID, prID, branchTag) {
...
  * @param {any} data file data
+ * @param {string} fileType Optional: file type so it knows where to save the file (default python)
+ * It's needed because go files should be in the GOPATH.
  */
-function saveFileToPRCache (repoID, prID, branchTag, filePath, data) {
-  writeFileCreateDirs(path.join('cache', repoID.toString(), prID.toString(), branchTag, filePath), data)
+function saveFileToPRCache (repoID, prID, branchTag, filePath, data, fileType) {
+  fileType = fileType || 'python'
+  if (fileType === 'go') {
+    writeFileCreateDirs(path.join('cache/go/src', repoID.toString(), prID.toString(), filePath), data)
```
evqna: Same here

evqna:
Alright, thanks for the writeup! Makes it much clearer.

PR #59, opened 2018-12-10
Fix the problem to use Git on Heroku

MVrachev:
I added a .profile file. In an articale from Heroku they explain what is the procfile used for:

"During startup, the container starts a bash shell that runs any code in $HOME/.profile before executing the dyno’s command."
From: https://devcenter.heroku.com/articles/dynos#startup 

First I create a bin folder (I can't upload it because it's in the .gitignore file) then I added two commands which are installing gosec.

Also I added a few lines in which I create an .env file necessary for probot. I fill this file with the enviormental variables needed by probot, those  environmental variables are on Heroku I and get them from there.

### DIFF
--- a/.profile
+++ b/.profile
@@ -1,0 +1,17 @@
+#!/bin/sh
+set -x
+
+mkdir $HOME/bin
+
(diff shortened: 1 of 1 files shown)


evqna:
Not sure about this, the `.env` file should not be necessary as the environment variables are already defined in the Heroku settings and available, there is no need to export them again.

PR #60, opened 2018-12-10
Fix a bug when call gosec with more than one go file

MVrachev:
There is a bug when we call gosec with multiple go files.

The reason I belive is that in the gosecArgs array when I give another array - goFiles as argument javascirpt somehow flattens the goFiles array and Gosec recives something strange as argument.

The reason I belive this is that when I concatenate the args array with the gofile array now it works with pull requests with multiple go files.

I added a test for this case.

### DIFF
--- a/gosec/gosec.js
+++ b/gosec/gosec.js
@@ -19,6 +19,7 @@ module.exports = (directory, inputFiles, reportFile) => {
   }
   reportFile = reportFile || 'gosec.json'
   const currentDirectory = path.join(__dirname, '../', directory)
+
   /**
   * @argument gosec command which the child process will execute
   * @argument -fmt output format of the command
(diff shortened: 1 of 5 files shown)


gosec/gosec.js
```diff
@@ -19,14 +19,16 @@ module.exports = (directory, inputFiles, reportFile) => {
...
   const currentDirectory = path.join(__dirname, '../', directory)
+
   /**
   * @argument gosec command which the child process will execute
   * @argument -fmt output format of the command
   * @argument out flag which redirects the gosec output to a file
   * @argument reportFile the file where the output of gosec will be stored
   * @argument goFiles files which will be analyzed by gosec
   */
-  let gosecArgs = ['-fmt=json', '-out', reportFile, goFiles]
```
evqna: The javascript spread operator `...` does exactly this (see the bandit runner).

evqna:
I thought I fixed this and a few similar nits, let me see if I can find it.

gosec/gosec.js
MVrachev: Yep, you were right. I change my code to use this operator.

PR #62, opened 2018-12-11
Relative file paths in Gosec issues

MVrachev:
Right now when there is a issue found by Gosec the path to the file will be the full path to the file in the machine where Precaution is working.

Here is an example: 
image

It's from this pull request: https://github.com/MVrachev/Travic-CI---Tests/pull/687/checks

Also as you can when you go to the 'Files changed' the issues are not shown there.

With this pull request I fix this with removing the unnecessary path from the issue.

Here is how it will be with this fix: 
image

It's from this pull request: https://github.com/MVrachev/Travic-CI---Tests/pull/695/checks

As you can see in 'Files changed' tab the issues are now shown right where the problematic code is.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,8 @@ const cache = require('./cache')
 const { config } = require('./config')
 const apiHelper = require('./github_api_helper')
 
+const path = require('path')
+
 /**
  * @param {import('probot').Application} app - Probot's Application class.
  */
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -81,7 +83,7 @@ async function runLinterFromPRData (pullRequests, context, headSha) {
     const banditReport = generateBanditReport(banditResults, cache.getBranchPath(repoID, PR.id, 'head', 'bandit'))
 
     const gosecResults = await runGosec(cache.getBranchPath(repoID, PR.id, 'head', 'gosec'), inputFiles)
-    const gosecReport = generateGosecReport(gosecResults)
+    const gosecReport = generateGosecReport(gosecResults, path.join(__dirname, cache.getBranchPath(repoID, PR.id, 'head', 'gosec')))
```
evqna: Prefer using `path.resolve` to convert relative paths to absolute ones rather than `__dirname`.

evqna: There are a few more instances of this in the code that should be changed too.

evqna:
Looks good, apart from the fact that your spelling of my name gets crazier every time 😉

Issue #64, opened 2018-12-11
Remove the issues found summary


ericwb:
**Describe the bug**
When running Precaution, the output shows:
SEVERITY_HIGH: 10
SEVERITY_MEDIUM: 8
SEVERITY_LOW: 0

These don't have much value to the end user since he/she doesn't know what a HIGH vs. LOW is in the output since the App only permits Error and Warning levels.  

**To Reproduce**
1. Install precaution on a repo
2. Create a PR with security issues
3. See output

**Expected behavior**
I don't expect those items to be listed.

**Screenshots**

**Additional context**

PR #71, opened 2018-12-13
New summary

MVrachev:
I changed our summary. The messages now are making more sense.

This is fix for issues https://github.com/vmware/precaution/issues/64
 
Because there are many possible situation I wanted to make the code smaller as possible.
I hope it's readable.

I had to change besides merge_reports also the bandit and gosec report functions so they will give summary relative to the new merged summary. Now I get the amount of errors, warnings and notices from the annotations of Gosec and Bandit not from the raw summary given by Gosec and Bandit.

I added a useful function with which I count the amount of errors, warnings and notices in annotation array in annotations_levels.

Also I had to change the tests in test/merge.reports.test.js so they won't fail.

### DIFF
--- a/annotations_levels.js
+++ b/annotations_levels.js
@@ -3,15 +3,19 @@
 
 const { annotationsLevels } = require('./config')
 
+/**
+ * @param {Object[]} annotations the issues found by a security linter wrapped into this object:
+ * https://developer.github.com/v3/checks/runs/#annotations-object
+ */
 function countAnnotationLevels (annotations) {
   let errors = 0
   let warnings = 0
   let notices = 0
 
-  for (let i = 0; i < annotations.length; ++i) {
-    if (annotations[i].annotation_level === 'failure') {
+  for (let annotation of annotations) {
+    if (annotation.annotation_level === 'failure') {
       errors += 1
-    } else if (annotations[i].annotation_level === 'warning') {
+    } else if (annotation.annotation_level === 'warning') {
       warnings += 1
     } else {
       notices += 1
--- a/annotations_levels.js
+++ b/annotations_levels.js
@@ -3,6 +3,23 @@
 
 const { annotationsLevels } = require('./config')
 
+function countAnnotationLevels (annotations) {
+  let errors = 0
+  let warnings = 0
+  let notices = 0
+
+  for (let i = 0; i < annotations.length; ++i) {
+    if (annotations[i].annotation_level === 'failure') {
+      errors += 1
+    } else if (annotations[i].annotation_level === 'warning') {
+      warnings += 1
+    } else {
+      notices += 1
+    }
+  }
+  return { errors, warnings, notices }
+}
+
 /**
  * @param {String} severity issue severity from bandit analyze
  * @param {String} confidence issue confidence from bandit analyze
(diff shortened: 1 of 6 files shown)


MVrachev:
Here are example messages:

image

when called like this:

image

ericwb:
I don't think stating what color a warning, error, or notice is necessary.  Plus, who knows, GitHub could change the icon and color on us.  Or if someone is color blind, it also wouldn't be relevant.

merge_reports.js
```diff
@@ -2,28 +2,42 @@
...
 const { config } = require('./config')
+/**
+ * @param {Number} errors the amount of errors found
+ * @param {Number} warnings the amount of warnings found
+ * @param {Number} notices the amount of notices found
+ */
+function getCorrectSummary (errors, warnings, notices) {
+  let summary = ''
+  let errorPrefix = errors > 1 ? 'There where ' + errors + ' errors found.' : 'There is 1 error found.'
+  let warningPrefix = warnings > 1 ? 'There where ' + warnings + ' warnings found.' : 'There is 1 warning found.'
```
ericwb:
```suggestion
  let warningPrefix = warnings > 1 ? 'There were ' + warnings + ' warnings found.' : 'There was 1 warning found.'
```

merge_reports.js
```diff
@@ -2,28 +2,42 @@
...
+  let noticePrefix = notices > 1 ? 'There where ' + notices + ' notices found.' : 'There is 1 notices.'
```
ericwb:
```suggestion
  let noticePrefix = notices > 1 ? 'There were ' + notices + ' notices found.' : 'There was 1 notices.'
```

annotations_levels.js
```diff
@@ -3,6 +3,23 @@
 
 const { annotationsLevels } = require('./config')
 
+function countAnnotationLevels (annotations) {
+  let errors = 0
+  let warnings = 0
+  let notices = 0
+
+  for (let i = 0; i < annotations.length; ++i) {
```
ericwb:
```suggestion
   for (annotation in annotations) {
```

annotations_levels.js
```diff
@@ -3,6 +3,23 @@
...
+    if (annotations[i].annotation_level === 'failure') {
```
ericwb:
```suggestion
     if (annotation.annotation_level === 'failure') {
```

annotations_levels.js
```diff
@@ -3,6 +3,23 @@
...
+      errors += 1
+    } else if (annotations[i].annotation_level === 'warning') {
```
ericwb:
```suggestion
     } else if (annotation.annotation_level === 'warning') {
```

evqna:
This still feels too wordy to me, I would rather simply have something such as: "Results: *n* Errors, *m* Warnings, *p* Notices".

MVrachev:
But that way there is clear separation from erros, notices and warnings.
Also that way we can put emojis.

evqna:
Well there's no need to have it all in one line, you can have a list like:
- ❌ 2 errors
- ⚠️ 5 warnings
- ℹ️ 3 notices
or whatever symbols you want to use

PR #72, opened 2018-12-14
Fix deletion

MVrachev:
I noticed that we delete the folder cache/prID/repoID not the whole cache/prID folder.

I we don't need to keep the pr-s subfolders.


evqna:
-1 for me, we may be processing multiple pull requests at the same time on one repository (if multiple branches are pushed to quickly or if several PRs are opened in a short time) and we definitely don't want one run of the app to delete the working directory of other runs.
This means we may leave behind an empty folder for every repo ID once every PR has been processed but that's not too big of an issue (and if it is we can fix that separately).

MVrachev:
A legitimate point. You are right.

Issue #57, opened 2018-12-08
Strange exception during tests


MVrachev:
**Describe the bug**
When I run 'yarn test' I get this strange exception:
![image](https://user-images.githubusercontent.com/16246778/49684886-b752a600-fae2-11e8-88a0-0c8ee427ffe5.png)

If you follow where it comes from it comes from the tests where we save gofile1.go, gofile2.go, pythonfile1.py and pythonfile2.py.

It looks like  gofile1.go are not loaded correctly gofile2.go

**To Reproduce**
1. Run yarn test on master branch on commit: 
07f5371f23ccf368e7b573f5aed757c3e4ea61f4
2. You will see the exception under the  PASS  test/index.test.js (5.937s) line.

**Expected behavior**
This exception to don't appear or the test to fail.

PR #73, opened 2018-12-17
Validation of the Go files

MVrachev:
This is a fix for this problem: https://github.com/vmware/precaution/issues/57

I wanted to make sure that Gosec will analyze valid Go files. The least thing I can do is to make sure that the file starts with a "package" statement as said here:
https://golang.org/doc/code.html#PackageNames

### DIFF
--- a/gosec/gosec.js
+++ b/gosec/gosec.js
@@ -6,6 +6,17 @@ const parse = require('../parse_output.js')
 const path = require('path')
 const fs = require('fs')
 
+function checkIfValidContent (pathToFile) {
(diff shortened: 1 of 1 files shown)


evqna:
I also don't see how that solves our problem, shouldn't we just make sure our test files are well formed?

MVrachev:
If we decide we don't have to check if the files are valid then we should make your test files: gofile1, gofile2 valid.
Antoine can you help with that?

evqna:
Sure, I will see what I can do.

evqna:
Solved in #76

Issue #65, opened 2018-12-11
Add links in the issues to reference docs


ericwb:
**Is your feature request related to a problem? Please describe.**
When running precaution, sometimes the user may not fully understand the issue being flagged.  That's where it would be very useful to have "more info" type links.

**Describe the solution you'd like**
For example:
```
 B413:blacklist

The pyCrypto library and its module DSA are no longer actively maintained and have been deprecated. Consider using pyca/cryptography library.
```
I'd like to see a link on B413:blacklist to the actual Bandit documentation for such error. Like:
https://bandit.readthedocs.io/en/latest/blacklists/index.html

In the output of Bandit, it will generate a `more_info` key that is an HTTP link to the reference documentation.  So we should be able to use that.  

Not sure if Gosec provides the same, but worth looking into.

**Describe alternatives you've considered**
none

**Additional context**
https://github.com/ericwb/bandit/pull/2/checks

MVrachev:
I created a pull request to add "more info" on the issues found by Bandit.

Gosec at least for now doesn't have a good documentation about the issues it finds.

Here is a thread about this topic in which I participated:
https://github.com/securego/gosec/issues/127

ericwb:
What about the `raw_details` field of the annotation object?

https://developer.github.com/v3/checks/runs/#annotations-object

MVrachev:
I tried with raw_details. First with markdown: 

![image](https://user-images.githubusercontent.com/16246778/50207237-90179680-0376-11e9-872d-b80b24db23cd.png)

Then with html: 
![image](https://user-images.githubusercontent.com/16246778/50207289-b50c0980-0376-11e9-891a-4430a8c1c25a.png)

This is how I put the html tag: 
![image](https://user-images.githubusercontent.com/16246778/50207305-bfc69e80-0376-11e9-8159-94d00fbb3e8f.png)

MVrachev:
I tried the same thing with message/title:

Message/title with markdown: 
![image](https://user-images.githubusercontent.com/16246778/50207551-79be0a80-0377-11e9-8a97-4b41a522936b.png)

Message/title HTML: 
![image](https://user-images.githubusercontent.com/16246778/50207718-eb965400-0377-11e9-84e6-eacea3c8eac6.png)

evqna:
@MVrachev The [checks run documentation](https://developer.github.com/v3/checks/runs/#output-object-1) says markdown is supported by the **summary** and **text** fields of the output object, did you try those?

[...]

PR #75, opened 2018-12-17
Add more info links on Bandit issues

MVrachev:
The json output of Bandit gives us a "more_info" weblink where there is more information about the issue.
Here is an example how it will work with this format of the issues found by Bandit:
https://github.com/MVrachev/Travic-CI---Tests/pull/724/checks?check_run_id=41872903

Gosec for now doesn't support such a thing. :(

Hopefully in future they will support this. I follow this thread for this topic:
https://github.com/securego/gosec/issues/127


MVrachev:
I will try to do that tomorrow.

MVrachev:
The title doesn't support markdown nor I had any success with html tags. :(

MVrachev:
As I mentioned here: 
https://github.com/vmware/precaution/issues/65#issuecomment-448909029
then we will have to put a link to the Bandit documentation as a whole.

PR #76, opened 2018-12-17
Simplify handling of fixture files in tests.

### LINKED ISSUE #57: Strange exception during tests

evqna:
Closes #57


evqna:
The PR fixtures should only reference files that are loaded in beforeAll otherwise attempts to access their contents will result in undefined.

evqna:
I might add an error check there to prevent future issues.

MVrachev:
You did some good work here but now we don't have a test where we download the files as before.
Shouldn't we add go files in the pullRequests.listFiles ?

evqna:
Should be good now.

PR #78, opened 2018-12-20
Linter refactor

evqna:
Currently the code that deals with running linters and collecting the results of analysis is scattered around the code base which makes it difficult to change and understand. Additionally, intricacies such as the way gosec interacts with the folder structure have to be accounted in various places which also increases the mental overhead when dealing with simple tasks.

This PR attempts to hide all linter-specific behavior under a common interface `Runner` that executes the common scanning and reporting logic. The `linters` folder contains the specifics of each linter runner instance (command line arguments, report parsing logic, ...). I'm not entirely satisfied with the current design (having `new Bandit()` in javascript is quite ugly IMO), i'm open to suggestions to change it.

Some other notable changes:
- I dropped the baseline code as it was specific to Bandit and we weren't using it anymore
- The runner only gives the GitHub report back, not the intermediate scan output so I had to modify all tests
 
TODO:
[x] Switch the main app logic to new runner code
[x] Update all tests to use new mechanism
[x] Document interface and new code
[ ] Figure a systematic way to log information
[ ] Merge report generation code with new linter file?


cache.js
```diff
@@ -35,19 +34,16 @@ function branchPathExists (repoID, prID, branchTag) {
...
  */
-function saveFileToPRCache (repoID, prID, branchTag, filePath, data, fileType) {
+function saveFileToPRCache (repoID, prID, filePath, data, fileType) {
   fileType = fileType || 'python'
-  if (fileType === 'go') {
-    writeFileCreateDirs(path.join('cache/go/src', repoID.toString(), prID.toString(), filePath), data)
-  } else if (fileType === 'python') {
-    writeFileCreateDirs(path.join('cache', repoID.toString(), prID.toString(), branchTag, filePath), data)
-  }
+  const appTag = (fileType === 'go') ? 'gosec' : 'bandit'
```
ericwb: Should probably start using a const for 'gosec' and 'bandit'.  I've seen the strings more than once.  Probably from bandit.js and gosec.js would be appropriate.

runner.js
```diff
@@ -0,0 +1,63 @@
...
+const Gosec = require('./linters/gosec')
+
+/**
+ * Run all linters on specified files
+ * @param {string[]} files Files to analyze
+ * @param {string} repoID
+ * @param {string} prID
+ */
+async function runLinters (files, repoID, prID) {
+  // TODO: Sync with file download location resolution
```
ericwb: Also in the future, it would be nicer if the linters were pluggable and not hardcoded here.  I imagine a linter registration model, that the runner would iterate through the loaded linters.

linters/bandit.js
```diff
@@ -0,0 +1,64 @@
+// Copyright 2018 VMware, Inc.
+// SPDX-License-Identifier: BSD-2-Clause
+
+const cache = require('../cache')
+
+const report = require('../bandit/bandit_report')
+
+module.exports = class Bandit {
```
ericwb: Does javascript have abstract classes or interfaces the class would define?

runner.js
```diff
@@ -0,0 +1,63 @@
...
+  // Promise report generation
+  return new Promise((resolve, reject) => {
+    process.on('error', reject)
+    process.on('close', () => reportHandler(linter, reportFilePath, resolve, reject))
+  })
+}
+
+function reportHandler (linter, reportFilePath, resolve, reject) {
+  fs.readFile(reportFilePath, (err, data) => {
+    if (err) { reject(err) } else {
```
ericwb:
```suggestion
    if (err) {
       reject(err)
     } else {
```

linters/bandit.js
```diff
@@ -0,0 +1,64 @@
...
+    return JSON.parse(data)
+  }
+
+  /**
+   * Generates a report in the format expected by GitHub checks
+   * from the linter results
+   * @param {any} results Linter results
+   * @returns GitHub checks report
+   */
+  generateReport (results) {
```
ericwb: Looks like both bandit.js and gosec.js have this method which just calls another.  Do we anticipate future linters doing other work here?

linters/bandit.js
```diff
@@ -0,0 +1,64 @@
...
+  args (files) {
+    return ['--format', 'json', '-o', this.reportPath, ...files]
+  }
+
+  /**
+   * Parses the linter results
+   * @param {Buffer} data The raw linter results data
+   */
+  parseResults (data) {
+    return JSON.parse(data)
```
ericwb: Looks like both bandit.js and gosec.js have this method which just JSON.parse()  Do we anticipate future linters doing other work here?

index.js
```diff
@@ -123,19 +100,14 @@ async function processPullRequest (pullRequest, context) {
...
-      const status = fileJSON.status
 
       const headRevision = apiHelper.getContents(context, filename, ref)
 
-      if (config.compareAgainstBaseline && status === 'modified') {
-        const baseRevision = apiHelper.getContents(context, filename, baseRef)
-        cache.saveFile(repoID, id, 'base', filename, (await baseRevision).data, 'python')
-      }
-
+      // TODO: merge this code with linter-specific path resolution
```
ericwb: Is this a TODO for a future PR or future commit on this PR?

linters/bandit.js
MVrachev: I thought the same. Do we need parseResults as a member function?

runner.js
MVrachev: I thought the same yes. That would be really cool.

test/gosec.spawn.test.js
```diff
@@ -1,36 +1,41 @@
...
 // SPDX-License-Identifier: BSD-2-Clause
 
-const gosec = require('../gosec/gosec')
 const fs = require('fs-extra')
 
-describe('Spawn gosec tests', () => {
-  test('Empty input for gosec', async () => {
-    const gosecResult = await gosec('', [])
-    expect(gosecResult).toBeFalsy()
-  })
```
MVrachev: Do we need to remove to remove the test were we give empty array as input file ?

test/gosec.spawn.test.js
```diff
@@ -1,36 +1,41 @@
...
-  test('Run gosec on go problematic file', async () => {
-    const gosecResult = await gosec('test/fixtures/go/src/bad_files', ['bad_test_file.go'])
-    expect(gosecResult.Stats.found).toBeGreaterThan(0)
-    fs.remove('test/fixtures/go/src/bad_files/gosec.json')
+  test('Handles packages with multiple files', async () => {
+    const report = await gosec('test/fixtures/go/src/multiple_bad_files/', ['bad_test_file.go', 'networking_binding_test.go', 'randNumTest.go'])
+
+    expect(report.annotations.length).toEqual(8)
+    expect(report.annotations[4].start_line).toEqual('9')
+    expect(report.annotations[7].start_line).toEqual('16')
```
MVrachev: Maybe it's a good idea to add a check how many files Gosec checked.
Because as you remembered if we place in this folder a file which ends with *_test.go then Gosec (because this is a Go rule) will think that this is a go test file and won't check it.
I mean this:
`` expect(gosecResult.Stats.files).toEqual(3)``

test/gosec.spawn.test.js
```diff
@@ -1,36 +1,41 @@
...
-  })
+const { run } = require('../runner')
+const Gosec = require('../linters/gosec')
+
+function gosec (dir, files) {
+  return run(new Gosec(), dir, files)
+}
+
+describe('Gosec runner', () => {
+  test('Finds issues in invalid file', async () => {
```
MVrachev: I don't think that "Finds issues in invalid file" is a good name for that test.
Because when you say invalid for me it means a file with non valid Go code which isn't compiling.

test/gosec.spawn.test.js
```diff
@@ -1,36 +1,41 @@
...
+    const report = await gosec('test/fixtures/go/src/bad_files', ['bad_test_file.go'])
 
-  test('Run gosec on non go files', async () => {
-    const gosecResult = await gosec('test/fixtures/go/src/non_go_files', ['hello_world.py'])
-    expect(gosecResult).toBeFalsy()
```
MVrachev: I think that it's a good idea to keep that test were we give a non go file to gosec.
Because whatever changes we are doing we should be aware we are filtering for non go files.
The problem is that Gosec is really sensitive and when you give Gosec non go files it throws an exception.

runner.js
evqna: Yes, this is a rough first draft but it would be nice to have an array of objects and just iterate through it. I'll see what I can do.

linters/bandit.js
evqna: I'm not very familiar with "design patterns" in javascript, I did some research on this and it appears there isn't (we would need to switch to typescript for this).
I might try another approach based on delegation (the prototypal equivalent of inheritance) but I don't know if it would solve the problem of enforcing a common interface for all linter objects.

linters/bandit.js
evqna: Maybe in the future we will stumble upon a linter that doesn't support JSON outputs so we will need manual parsing logic. Pretty unlikely but it doesn't cost much.

evqna:
@MVrachev Yes I didn't put the logging code back in yet because I wanted to try to use Bunyan (the probot logger) instead of `console.log` but all things considered this is outside the scope of this PR. I'll copy your changes and we can decide on the logging situation later.

test/gosec.spawn.test.js
evqna: I'll change the title

linters/bandit.js
evqna: For now the report generation logic lives in a separate file for bandit and gosec but it may not be necessary as most of it is similar and could be factored away.
This PR is already large enough so right now I'm taking the simplest approach.

test/gosec.spawn.test.js
evqna: What title did you have in mind?

test/gosec.spawn.test.js
evqna: Fixed.

test/gosec.spawn.test.js
evqna: I'm not sure there is much we can do about this as we don't specify individual go files to scan, we just give it a directory and let gosec scan it recursively.
The logic to save go files to a specific directory is not part of the runner code so it doesn't make sense to test it here.

evqna:
Ready for review.

test/gosec.spawn.test.js
MVrachev: "Find issues in non secure file" or
"Find issues in vulnerable file" or
"Find issues in problematic file"
If change that maybe you should also rename the mix.invalid file.

test/gosec.spawn.test.js
MVrachev: I don't think we should download the go files in other special folder.
I think we should keep the structure of the project and pr.
We know if there are any go file for scan because we filter them and do that check:
image

test/gosec.spawn.test.js
MVrachev: I still think that check won't hurt.

test/gosec.spawn.test.js
evqna: Switched to 'vulnerable'.

test/gosec.spawn.test.js
evqna: Well we are doing this because the go code needs to be in the GOPATH which is currently set to cache/go.

test/gosec.spawn.test.js
evqna: I'm not sure what you mean exactly, the fact that gosec ignores these files is outside of our control so I don't see the need to test it.

linters/bandit.js
```diff
@@ -0,0 +1,68 @@
...
+  get name () {
+    return 'bandit'
+  }
+
+  /**
+   * The path to the generated report file
+   * relative to the working directory
+   */
+  get reportPath () {
+    return '../bandit.json'
```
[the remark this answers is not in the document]
evqna: Looking into it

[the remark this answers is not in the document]
MVrachev: Well I don't see change here.
Do you have other idea in mind or?

linters/gosec.js
```diff
@@ -0,0 +1,68 @@
...
+  get name () {
+    return 'gosec'
+  }
+
+  /**
+   * The path to the generated report file
+   * relative to the working directory
+   */
+  get reportPath () {
+    return '../gosec.json'
```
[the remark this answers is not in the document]
MVrachev: Again I expected change here. 
If you have different opinion I will be glad to hear it.

linters/index.js
```diff
@@ -0,0 +1,10 @@
+// Copyright 2018 VMware, Inc.
```
MVrachev: It's little vague for me why do we need this file: linters/index.js?
Also if this is a config file shouldn't the name be something like: config_linters or just config?

MVrachev: So when you include the linters folder in the runner file you basically include linters/index.js?

test/gosec.spawn.test.js
MVrachev: Do we check if there are any go files when we give the folder for scan to Gosec?
I think we do with the filter function?
Because this is really important because Gosec throws an error if there are no go files in the folder.

cache.js
evqna: I agree but I'd rather tackle that as a follow up.
Opening #103

runner.js
evqna: `linters/index.js` now exports a list of code linters, we might add a utility function to register linters dynamically later

evqna: Note that this requires the report generation rewrite, coming in a followup PR

linters/bandit.js
evqna: See #102 for discussion

linters/bandit.js
evqna: Also bringing everything in the same file might make it more difficult to test correctly, though I am open to suggestions.
Of course we can export every helper function but then there is not much point in getting rid of the helper files.

index.js
evqna: Future PR

test/gosec.spawn.test.js
evqna: If there are no go files the runner returns early and generates a default report.

linters/bandit.js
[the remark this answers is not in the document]
evqna: I don't think the linter specific code should concern itself with the save location of the report path, this should entirely be controlled by the calling code (the app batch runner or a unit test file). I might actually drop the '../' prefix and move that elsewhere.
The only current use is factoring out the report path between the args of the spawned linter and the callback that reads it back.

linters/gosec.js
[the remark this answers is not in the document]
evqna: Same here

linters/index.js
evqna: This decouples the runner code from specific linters: all that is necessary to enable or disable specific linters is to modify the fields in this object.

evqna: > So when you include the linters folder in the runner file you basically include linters/index.js?

Yes, importing a folder imports the `index.js` file if it exists.

test/gosec.spawn.test.js
MVrachev: Yep, you are right.

linters/index.js
MVrachev: But why the name of the file is index.js?
I think it would be more clearer if it is something like the above suggestions I gave.

Maybe this naming is some kind of JavaScript pattern but I don't know.

linters/bandit.js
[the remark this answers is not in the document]
MVrachev: I agree that the caller should specify where to save the report.
Maybe you can do something like
``get reportPath (workingDirectory) { return path.join(workingDirectory, '../bandit.json') }``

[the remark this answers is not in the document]
evqna: I'll see what I can do (it's tricky because getters cannot have parameters: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)

linters/index.js
evqna: So we can just write `requires('./linters')` instead of `requires('./linters/default_linters')` (makes more sense to me)

linters/bandit.js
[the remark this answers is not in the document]
MVrachev: I am sorry. You really don't need to give workingDirectory to this get function.
Maybe you should rename the function and use it like this:
``get reportFileName () { return 'bandit.json') }``

linters/index.js
MVrachev: Ookay, maybe you are right. It's not such a big deal. :D

linters/bandit.js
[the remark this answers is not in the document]
evqna: Done

PR #79, opened 2018-12-20
New summary

MVrachev:
I changed our summary. The messages now are making more sense.

Closes: https://github.com/vmware/precaution/issues/64

In this summary the annotations:
- with level 'failure' are errors
- with level 'warning' are warnings
- with level 'notice' are notices

Also I added a "text" field in the output object.
There I used that Bandit provides "more info" field for its issues. In the "more info" there is a link leading to the Bandit documentation of the issue.

Sadly Gosec doesn't support that now. :(

Here is an example how the new summary will look: 
image

It is from this example: 
https://github.com/MVrachev/Travic-CI---Tests/pull/772/checks?check_run_id=43369277

### DIFF
--- a/bandit/bandit_report.js
+++ b/bandit/bandit_report.js
@@ -10,7 +10,7 @@ const { countIssueLevels } = require('../annotations_levels')
  */
 function createMoreInfoLinks (issues) {
   let issuesMap = new Map()
-  let moreInfo = 'For more information about the issues refer to the following: \n'
+  let moreInfo = 'For more information about the issues follow the links: \n'
 
   for (let issue of issues) {
     if (issuesMap.has(issue.test_id) === false) {
@@ -23,31 +23,19 @@ function createMoreInfoLinks (issues) {
 }
 
 /**
- * @param {*} annotations all issues found by Bandit wrapped in the annotation object
- * see: https://developer.github.com/v3/checks/runs/#annotations-object
- */
-function customSummary (annotations) {
-  const { errors, warnings, notices } = countIssueLevels(annotations)
-
-  const summary = {
-    'errors': errors,
-    'warnings': warnings,
-    'notices': notices
-  }
-  return summary
-}
-
-/**
  * Convert bandit output into valid 'output' object for check run conclusion
  * @param {any} results Bandit json output
  */
 module.exports = (results) => {
-  let title, summary, annotations, moreInfo
+  let title = ''
+  let summary = ''
+  let annotations = []
+  let moreInfo = ''
 
   if (results && results.results.length !== 0) {
     title = config.issuesFoundResultTitle
     annotations = results.results.map(issue => getAnnotation(issue))
-    summary = customSummary(annotations)
+    summary = countIssueLevels(annotations)
     moreInfo = createMoreInfoLinks(results.results)
   } else {
     title = config.noIssuesResultTitle
--- a/bandit/bandit_report.js
+++ b/bandit/bandit_report.js
@@ -10,7 +10,7 @@ const { countIssueLevels } = require('../annotations_levels')
  */
 function createMoreInfoLinks (issues) {
   let issuesMap = new Map()
-  let moreInfo = 'For more information about the issues follow the links: \n'
+  let moreInfo = 'For more information about the issues refer to the following: \n'
 
   for (let issue of issues) {
     if (issuesMap.has(issue.test_id) === false) {
--- a/bandit/bandit_report.js
+++ b/bandit/bandit_report.js
@@ -3,16 +3,36 @@
 
 const { config } = require('../config')
 const { getAnnotation } = require('./bandit_annotations')
+const { countIssueLevels } = require('../annotations_levels')
 
-function customSummary (banditSummary) {
-  let severityHigh = banditSummary['SEVERITY.HIGH']
-  let severityMedium = banditSummary['SEVERITY.MEDIUM']
-  let severityLow = banditSummary['SEVERITY.LOW']
+/**
+ * @param {*} issues the issues found by Bandit
+ */
+function createMoreInfoLinks (issues) {
+  let issuesMap = new Map()
+  let moreInfo = 'For more information about the issues follow the links: \n'
+
+  for (let issue of issues) {
+    if (issuesMap.has(issue.test_id) === false) {
+      issuesMap.set(issue.test_id)
+      const text = `${issue.test_id}:${issue.test_name}`
+      moreInfo += `[${text}](${issue.more_info})\n`
+    }
+  }
+  return moreInfo
+}
+
+/**
+ * @param {*} annotations all issues found by Bandit wrapped in the annotation object
+ * see: https://developer.github.com/v3/checks/runs/#annotations-object
+ */
+function customSummary (annotations) {
+  const { errors, warnings, notices } = countIssueLevels(annotations)
 
   const summary = {
-    'SEVERITY_HIGH': severityHigh,
-    'SEVERITY_MEDIUM': severityMedium,
-    'SEVERITY_LOW': severityLow
+    'errors': errors,
+    'warnings': warnings,
+    'notices': notices
   }
   return summary
 }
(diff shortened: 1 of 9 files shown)


bandit/bandit_report.js
```diff
@@ -3,16 +3,36 @@
...
+    }
+  }
+  return moreInfo
+}
+
+/**
+ * @param {*} annotations all issues found by Bandit wrapped in the annotation object
+ * see: https://developer.github.com/v3/checks/runs/#annotations-object
+ */
+function customSummary (annotations) {
```
evqna: This function is an extra layer of indirection that does nothing: it extracts the returned values from `countIssueLevels` and rebinds them to the same names.
You can drop it and call `countIssueLevels` directly.

bandit/bandit_report.js
```diff
@@ -22,15 +42,16 @@ function customSummary (banditSummary) {
...
  */
 module.exports = (results) => {
-  let title, summary, annotations
+  let title, summary, annotations, moreInfo
 
   if (results && results.results.length !== 0) {
     title = config.issuesFoundResultTitle
-    summary = customSummary(results.metrics._totals)
     annotations = results.results.map(issue => getAnnotation(issue))
+    summary = customSummary(annotations)
```
evqna:
```suggestion
    summary = countIssueLevels(annotations)
```

merge_reports.js
```diff
@@ -3,27 +3,42 @@
 
 const { config } = require('./config')
 
+/**
```
evqna: One line of explanation would be nice here (What does "correct summary" mean?)

merge_reports.js
```diff
@@ -50,5 +66,5 @@ module.exports = (banditReport, gosecReport) => {
   } else {
     annotations = annotations.concat(gosecReport.annotations, banditReport.annotations)
   }
-  return { title, summary, annotations }
+  return { title, summary, annotations, text }
```
evqna: I don't really like the idea of having `undefined` variables even though it works out in this case (the GitHub API will happily ignore it). Is there another way?

test/merge.reports.test.js
```diff
@@ -24,8 +24,9 @@ describe('Merge reports tests from Bandit and Gosec reports', () => {
     let banditReport = { title: config.issuesFoundResultTitle, summary: banditSummary, annotations: banditAnnotations }
     let gosecReport = { title: config.noIssuesResultTitle, summary: config.noIssuesResultSummary, annotations: [] }
 
-    const expectedSummary = 'SEVERITY_HIGH: ' + banditSummary.SEVERITY_HIGH + '\n' + 'SEVERITY_MEDIUM: ' +
-      banditSummary.SEVERITY_MEDIUM + '\n' + 'SEVERITY_LOW: ' + banditSummary.SEVERITY_LOW + '\n'
+    let expectedSummary = ':x: 1 error\n' +
```
evqna: shouldn't this be a const?

bandit/bandit_report.js
```diff
@@ -3,16 +3,36 @@
...
-function customSummary (banditSummary) {
-  let severityHigh = banditSummary['SEVERITY.HIGH']
-  let severityMedium = banditSummary['SEVERITY.MEDIUM']
-  let severityLow = banditSummary['SEVERITY.LOW']
+/**
+ * @param {*} issues the issues found by Bandit
+ */
+function createMoreInfoLinks (issues) {
+  let issuesMap = new Map()
+  let moreInfo = 'For more information about the issues follow the links: \n'
```
ericwb:
```suggestion
   let moreInfo = 'For more information about the issues refer to the following: \n'
```

merge_reports.js
```diff
@@ -3,27 +3,42 @@
 
 const { config } = require('./config')
 
+/**
+ * @param {Number} errors the number of errors found
+ * @param {Number} warnings the number of warnings found
+ * @param {Number} notices the number of notices found
+ */
+function getCorrectSummary (errors, warnings, notices) {
```
ericwb: What is a "correct" summary?  Versus an "incorrect" summary?

merge_reports.js
MVrachev: Yes, both of you with Eric are right. I will fix it.

merge_reports.js
MVrachev: Yes, both of you with Antoine are right. I will fix it.

test/merge.reports.test.js
MVrachev: Yes, I fixed it.

bandit/bandit_report.js
MVrachev: Yep, you are right.

merge_reports.js
MVrachev: The other two ways is to return in the if/else statements but that's something Eric told me long ago that he considers for not a good style.
Right @ericwb ? 

The other way is to check if the value is undefined in the end when you return which is little redundant. 

I used the current method on many occasions but I will change that just so we can see the alternative. I agree with you, you have a point.

evqna: I suggest you initialize `text` as an empty string, that way it is functionally identical and we don't have to worry about possible undefined values.

gosec/gosec_report.js
```diff
@@ -34,12 +14,12 @@ module.exports = (results, directory) => {
   let title, summary, annotations
   if (results && results.Issues.length !== 0) {
     title = config.issuesFoundResultTitle
-    summary = customSummary(results.Issues)
     annotations = results.Issues.map(issue => getAnnotation(issue, directory))
-  } else {
-    title = config.noIssuesResultTitle
-    summary = config.noIssuesResultSummary
+    summary = countIssueLevels(annotations)
+    return { title, summary, annotations }
```
evqna: Returning different objects at different places in a function causes the same problem (the caller will get a bunch of undefined if the function returns fewer values than expected).
Here I think we should always return the same structure (title + summary + annotations) and pick sensible defaults for the values that are not needed (empty string, empty array, ...)

evqna:
Other than that changes look good 👍

gosec/gosec_report.js
MVrachev: Yes, you are right. Your suggestion makes more sense.

merge_reports.js
MVrachev: Yes, you are right.

merge_reports.js
```diff
@@ -32,23 +49,27 @@ function mergeSummaries (banditSummary, gosecSummary) {
...
-  let title, summary
+  let title = ''
+  let summary = ''
+  let text = ''
   let annotations = []
 
   if (banditReport.title === config.noIssuesResultTitle && banditReport.title === gosecReport.title) {
     title = config.noIssuesResultTitle
     summary = config.noIssuesResultSummary
-  } else {
```
evqna: This should be reverted (put back in else block so there is only one return statement at the end of the function)

merge_reports.js
```diff
@@ -3,27 +3,44 @@
 
 const { config } = require('./config')
 
+/**
+ * Creates the final summary based on how many errors, warnings and notices there is
```
evqna:
```suggestion
 * Creates the final summary message describing the number of errors, warnings and notices.
```

merge_reports.js
```diff
@@ -3,27 +3,44 @@
...
+ * so the final summary to be meaningful.
+ * @param {Number} errors the number of errors found
+ * @param {Number} warnings the number of warnings found
+ * @param {Number} notices the number of notices found
+ */
+function createFinalSummaryMessage (errors, warnings, notices) {
+  let summary = ''
+
+  const errorsMessage = errors > 1 ? ':x: ' + errors + ' errors\n'
```
evqna:
```suggestion
  const errorsMessage = `:x: ${errors} errors\n`
```

MVrachev: If you use `:x: ${errors} errors\n` only you will have a situation when the output will be:
`1 errors` which we don't want.

merge_reports.js
MVrachev: I agree. I will fix it.

merge_reports.js
```diff
@@ -4,7 +4,7 @@
 const { config } = require('./config')
 
 /**
- * Creates the final summary based on how many errors, warnings and notices there is
+ * Creates the final summary message describing the number of errors, warnings and notices.
  * so the final summary to be meaningful.
```
evqna: This line can be removed

PR #82, opened 2018-12-21
Use if exists the Bandit config file

MVrachev:
If the user creates a ".bandit" file in this format:
https://github.com/PyCQA/bandit#configuration
we would use those options when we run Bandit.

An example of a ".bandit" configuration file:
https://docs.openstack.org/bandit/latest/config.html

P.S. I am searching for a way to define a gosec config file but is little unclear for me.

### DIFF
--- a/bandit/bandit.js
+++ b/bandit/bandit.js
@@ -4,6 +4,7 @@
 const { spawn } = require('child_process')
 const path = require('path')
 const parseOutput = require('../parse_output')
+const fs = require('fs')
 
 /**
  * Spawn a bandit process analyzing all given files in provided directory
@@ -25,6 +26,11 @@ module.exports = (directory, inputFiles, params) => {
   const reportPath = path.join(directory, params.reportFile)
 
   let banditArgs = [...pyFiles, '--format', 'json', '-o', params.reportFile]
+
+  if (fs.existsSync(path.resolve('.bandit'))) {
+    banditArgs.push('-c', path.resolve('.bandit'))
+  }
+
   if (params.baselineFile) {
     banditArgs.push('--baseline', params.baselineFile)
   }


MVrachev:
Here is an example how it works:

I have this .bandit file: 
bandit

I run it locally with those changes (my app is called MVrachevApp) and also I run it with the last version of the master branch and here are the results: 
https://github.com/MVrachev/Travic-CI---Tests/pull/781/checks?check_run_id=43637833

There are no problems found by my app because I ignore rule B309 and the file I analyze has problems only connected with this rule but the current official release finds those mistakes.  

The reason why you still see the problems underneath the code lines in the "Files changed" tab is that the official app (not "MVrachevApp") finds the issues connected with rule B309.

ericwb:
My concern with using the bandit config mechanism is that we will have an issue where there will be conflicts.  A .precaution.yml file will state one thing to skip, while a bandit config will state not to skip.  Which config would take precedence?  That might become confusing for users.

bandit/bandit.js
```diff
@@ -25,6 +26,11 @@ module.exports = (directory, inputFiles, params) => {
   const reportPath = path.join(directory, params.reportFile)
 
   let banditArgs = [...pyFiles, '--format', 'json', '-o', params.reportFile]
+
+  if (fs.existsSync(path.resolve('.bandit'))) {
+    banditArgs.push('-c', path.resolve('.bandit'))
```
ericwb: This can cause problems.  A .bandit file permits setting the targets (files), which would conflict with what we're trying to scan in the PR.

```
    ini_options = _get_options_from_ini(args.ini_path, args.targets)
    if ini_options:
        # prefer command line, then ini file
        args.excluded_paths = _log_option_source(args.excluded_paths,
                                                 ini_options.get('exclude'),
                                                 'excluded paths')

        args.skips = _log_option_source(args.skips, ini_options.get('skips'),
                                        'skipped tests')

        args.tests = _log_option_source(args.tests, ini_options.get('tests'),
                                        'selected tests')
        ini_targets = ini_options.get('targets')
        if ini_targets:
            ini_targets = ini_targets.split(',')
        args.targets = _log_option_source(args.targets, ini_targets,
                                          'selected targets')
```

MVrachev:
Yes, you are right. I will think of another way to do this.

Issue #67, opened 2018-12-13
Create a settings config file with Linter specific options


ericwb:
**Is your feature request related to a problem? Please describe.**
Some exploiters may want to pick and choose which checks are relevant to their project.  Bandit and Gosec do have ways to choose what types of checks/rules to include/exclude. We need to expose something like this in Precaution.

**Describe the solution you'd like**
Projects can create a .precaution.yml file to define which checks to run.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

MVrachev:
The problem is that there every user of Precaution has different app installation and how the user will have access to this .precaution.yml file?

We need an interface from where the user can access his .precaution.yml file and we need a way to host all of the .precaution.yml files.

ericwb:
The .precaution.yml file would be found in the target project repo.  This is easier than implementing a UI settings page.

MVrachev:
I think we should first merge the Antoines "Linter refactor" pull request and we can discuss what options we would give to the users with our precaution.ymal file.

MVrachev:
We (the Precaution maintainers) decided that we won't support linter specific options.

As we described it in `docs/configuration.md` file in pr https://github.com/vmware/precaution/pull/222
```
Precaution uses a variety of linters under the hood without exposing
implementation details and complexity of the underlying tools to the user. Because of that,
and in order to prevent conflict with any configuration files which may exist in the target git repository,
Precaution ignores any tool specific configuration
files which may exist in the project's repository and performs scans based solely
on Precaution specific configuration.
```

PR #83, opened 2018-12-27
Add precaution config file

MVrachev:
This pr is about this issues: https://github.com/vmware/precaution/issues/67 

We will need one more dependency: https://github.com/nodeca/js-yaml
It is not that easy to read properly a ymal file.
If you read it with fs.readFileSync you will have the content of a file as a string, not with objects which has properties.
Many people give an advice to use this library when dealing with ymal files.

I think that we can read for "skip_linters" key in our ymal file. With skip_linters the user can say which linters he want to skip when using Precaution.
Maybe it will be a good idea to give linter specific options. Like those for Bandit:
https://github.com/PyCQA/bandit#configuration

### DIFF
--- a/.precaution.ymal.example
+++ b/.precaution.ymal.example
@@ -1,0 +1,13 @@
+# Copyright 2018 VMware, Inc.
+# SPDX-License-Identifier: BSD-2-Clause
+
+# That way you can say which security linters do you want to skip
+skip_linters: 'bandit'
+
+# Example of Bandit specific options
+# All of them can be found here: https://github.com/PyCQA/bandit#configuration
+bandit:
+ targets: './test'
+ exclude: './cache'
+ skips: 'B101'
+ tests: ['B309', 'B203']
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "fs-extra": "^7.0.0",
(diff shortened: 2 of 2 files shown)


.precaution.ymal.example
```diff
@@ -0,0 +1,13 @@
+# Copyright 2018 VMware, Inc.
+# SPDX-License-Identifier: BSD-2-Clause
+
+# That way you can say which security linters do you want to skip
+skip_linters: 'bandit'
```
ericwb: This makes more sense when we have a plugin model for the linters.  Right now, we have just gosec and bandit which are hardcoded.

.precaution.ymal.example
```diff
@@ -0,0 +1,13 @@
...
+
+# Example of Bandit specific options
+# All of them can be found here: https://github.com/PyCQA/bandit#configuration
+bandit:
```
ericwb: I think we should attempt to abstract this more.  If we just blatantly expose bandit config parameters, then why have another file.  Especially if all the settings are the same.

MVrachev:
Well, then we can merge the Antoines "Linter refactor" pull request and we can discuss what options we would give to the users.
I wanted to bring a discussion with this pull request.

ericwb:
The discussion of the design should be in the Issue #83, not in the PR.

PR #89, opened 2019-01-02
Add coverage test runner and code coverage badge

MVrachev:
I use jest and Codecov tool (see codecov.io) to generate a badge about our projects code coverage.
I have to call yarn test which calls jest explicitly if I want to use this tool.
Look here: [CodeCov Node js examples](https://github.com/codecov/example-node#jest).

Codecov use Travis CI to run tests and gather information about the code coverage.

I tested and this badge is dynamically updated on every commit.

You can see how it looks here:
https://github.com/securego/gosec

This badge is configured for the master branch of precaution and that's why you can't see how it looks on my fork.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,11 @@ install:
   - pip3 install --user -r requirements.txt
   - curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $GOPATH/bin 1.2.0
   - export GOPATH=$GOPATH:${TRAVIS_BUILD_DIR}/test/fixtures/go
+  - npm install -g codecov
+
+script:
+   - yarn test
+   - codecov
 
 language: node_js
 node_js:
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 # Precaution
 
 [![Build Status](https://travis-ci.com/vmware/precaution.svg?branch=master)](https://travis-ci.com/vmware/precaution)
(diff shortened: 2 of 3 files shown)


ericwb:
The title of this PR is misleading.  It implies only a badge is added, but you are also adding a coverage test runner.

.travis.yml
```diff
@@ -8,6 +8,11 @@ install:
   - pip3 install --user -r requirements.txt
   - curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $GOPATH/bin 1.2.0
   - export GOPATH=$GOPATH:${TRAVIS_BUILD_DIR}/test/fixtures/go
+  - npm install -g codecov
+
+script:
+  - jest && codecov
```
ericwb: The default for script on node_js is `yarn test`.  Now you are switching to `jest`.

.travis.yml
```diff
@@ -8,6 +8,11 @@ install:
...
+  - codecov
```
ericwb: Why is `codecov` executed twice?

MVrachev:
I will test a little more what scripts do we need but I saw that that codecov needs their script so they can upload the test data to their website.

MVrachev:
In the script section of the ".travis.yml" file the codecov is absolutely necessary in order for codecov to upload the test and code coverage data to their website.

Then I tried to see if the code coverage data will be updated with only the codecov script but I it didn't worked. Even through Travis CI doesn't need explicitly to call the "yarn test" script to start the tests, Codecov needs that.

PR #90, opened 2019-01-08
Adding VSCode debug configurations

evqna:
After struggling for a while with setting up proper debugging for the app, I thought it would be useful to share my setup.
The outputCapture option is necessary to catch the logging output from Bunyan in the internal debugging console, which makes it easier to spot potential pitfalls not covered by our test assertions.


MVrachev:
I think this can be useful for future contributors.

MVrachev:
But maybe wait for Erics opinion.

docs/localdev.md
```diff
@@ -0,0 +1,9 @@
+<!--
+    Copyright 2018 VMware, Inc.
```
ericwb:
```suggestion
    Copyright 2019 VMware, Inc.
```

evqna: Looks like I still need to get used to it!

.vscode/launch.json
```diff
@@ -0,0 +1,30 @@
...
+            "type": "node",
+            "request": "launch",
+            "name": "Probot debug",
+            "program": "${workspaceFolder}/node_modules/probot/bin/probot-run.js",
+            "args": [
+                "./index.js"
+            ]
+        }
+    ]
+}
```
ericwb: sorry, another nit.  Can you add new line

Issue #80, opened 2018-12-21
The Re-run-all button doesn't work


MVrachev:
**Describe the bug**
When you have opened pull request and you have a scan of the pull request there is a "Re-run-all" button.
When you click it it just "Queued" and nothing happens.
![image](https://user-images.githubusercontent.com/16246778/50338123-0dc8d700-051c-11e9-845f-755c3f81c078.png)

**To Reproduce**
1. Create a pull request.
2. Wait for the first scan from Precaution.
3. Go to the "Checks tab".
4. On the left just next to the "Precaution" check there is this "Re-run-all" button. Click it.
5. See that nothing is happening.

**Expected behavior**
To re-run the Precaution test.

**Additional context**
I am not sure if this is a Precaution problem or Github problem. 
Worth investigating.

PR #91, opened 2019-01-09
Fix for rerequest button doesn't work

MVrachev:
Closes: https://github.com/vmware/precaution/issues/80

The problem was that when the user clicked the "Re-run" button then there would be created a check run rerequest even instead of a check suite rerequest event.

image

I added a small code that catches that case and I created a new unit test in index.test.js and a new file check_run_rerequested.json.

The file was needed to check if we catch the check run rerequest event correctly.


evqna:
Good catch, I didn't know this would send a check run event.

PR #92, opened 2019-01-09
Remove all baseline related code

evqna:
The baseline feature is now irrelevant since we rely on GitHub to correctly filter the annotations and only display relevant ones.

This is split off from #78 to help trim it down.


MVrachev:
Well Gosec doesn't have a baseline option so yes I too think this is unnecessary

PR #94, opened 2019-01-11
script: run standard before test suite

evqna:
Makes it easier to catch style errors quickly even when tests fail.


MVrachev:
A good point.

PR #98, opened 2019-01-14
Create a question template

MVrachev:
I think we should somehow do a more welcoming issue template for all users who just has questions for us.

That way we would encourage the users to ask more questions and we will have more and better feedback.


MVrachev:
Yes, but once we answer a question they will stay on Github for reference.
Also during the discussion it will be easier for others to participate.

ericwb:
Slack is a better messaging system than GitHub issues.  Issues I've seen on other repos that turn into a discussion get really hard to follow and a mess.  Similar to an email chain that gets replied, forwarded, but many people.

MVrachev:
Okay, I will create then pull request adding "Questions" section.

Issue #111, opened 2019-01-17
No commit found message


ericwb:
**Describe the bug**
Precaution erroring on PR of wavefront-adapter-for-istio

**To Reproduce**
1. do a dummy PR on wavefront-adapter-for-istio
2. Precaution check errors stating it couldn't find a commit
3. Then when commit is force-pushed, Precaution disappears from the listed checks
4. Reopening the PR, then Precaution reappears.

**Expected behavior**
Shouldn't see a No commit error when there is a commit.  Precaution shouldn't disappear from the listed checks.

**Screenshots**
https://github.com/vmware/wavefront-adapter-for-istio/pull/67/checks?check_run_id=51530610

**Additional context**
n/a

MVrachev:
It is confirmed that the problem is when you create a pull request across forks.

PR #113, opened 2019-01-18
Fix for "across the forks" PRs problems

MVrachev:
Close: https://github.com/vmware/precaution/issues/111

One of the bugs is when you make a pull request across forks. There was an error because it searched the files from the pr in the base repository instead of the forked repository.

The second bug is when there is a pull request across forks and then when you add commit to an existing pull request we didn't catch the
pull_request.synchronize event and then Precaution will queued for ever:
image

NOTE: The "Re-run" button doesn't work when there is pr across forks. The problem is that GitHub API give to Precaution an empty pull request object... I think this is a bug in the GitHub API and I will contact with them.

I wanted to publish this PR as fast as I could and that is why I didn't add unit tests for across the forks PRs.

### DIFF
--- a/github_api_helper.js
+++ b/github_api_helper.js
@@ -46,20 +46,20 @@ function getPRFiles (context, number) {
 /**
  * Get file contents as raw data
  * @param {import('probot').Context} context Probot context
- *  @param {any[]} pullRequests object which contains information about the pull request
  * @param {string} path file path relative to repository root
  * @param {string} ref sha of file revision
+ * @param {pull_request.head?} head Reference to the fork the commit originated from
  * @returns {Promise<any>} GitHub response
  * See https://developer.github.com/v3/repos/contents/#get-contents
  */
-function getRawFileContents (context, pullRequest, path, ref) {
+function getRawFileContents (context, path, ref, head) {
   let { owner, repo } = context.repo()
   // This check is necessary in the case when there is a pr
   // which is not from forked repository.
-  // Then the repo and owner fields are not in the pullRequest object
-  if (pullRequest.head.user) {
-    owner = pullRequest.head.user.login
-    repo = pullRequest.head.repo.name
+  // Then the repo and owner fields are not in the head object
+  if (head.user) {
+    owner = head.user.login
+    repo = head.repo.name
   }
 
   return context.github.repos.getContents({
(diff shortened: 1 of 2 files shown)


github_api_helper.js
```diff
@@ -46,13 +46,21 @@ function getPRFiles (context, number) {
...
  * @param {import('probot').Context} context Probot context
+ *  @param {any[]} pullRequests object which contains information about the pull request
  * @param {string} path file path relative to repository root
  * @param {string} ref sha of file revision
  * @returns {Promise<any>} GitHub response
  * See https://developer.github.com/v3/repos/contents/#get-contents
  */
-function getRawFileContents (context, path, ref) {
-  const { owner, repo } = context.repo()
+function getRawFileContents (context, pullRequest, path, ref) {
```
evqna: Rather than passing in the full PR object, add an optional head parameter so that the calling code can use the default behavior or opt-in to select a different repository than the one that sent the action.

```suggestion
function getRawFileContents (context, path, ref, head) {
```

github_api_helper.js
```diff
@@ -46,13 +46,21 @@ function getPRFiles (context, number) {
...
+  let { owner, repo } = context.repo()
+  // This check is necessary in the case when there is a pr
+  // which is not from forked repository.
+  // Then the repo and owner fields are not in the pullRequest object
```
evqna: Do not explain this inside the function, add it to the parameter JSdoc
ie `@param {pull_request.head?} head Reference to the fork the commit originated from`

index.js
```diff
@@ -111,7 +110,7 @@ async function processPullRequest (pullRequest, context) {
     .map(async fileJSON => {
       const filename = fileJSON.filename
 
-      const headRevision = apiHelper.getContents(context, filename, ref)
+      const headRevision = apiHelper.getContents(context, pullRequest, filename, ref)
```
evqna:
```suggestion
      const headRevision = apiHelper.getContents(context, filename, ref, pull_request.head)
```

github_api_helper.js
MVrachev: I still think this comment is useful but still I will add head as a parameter.

github_api_helper.js
MVrachev: Agree.

index.js
MVrachev: Done.

PR #117, opened 2019-01-22
Add integration test for success conclusion

evqna:
This should make it easier to spot issues with runs returning an error for the wrong reasons (problems with the environment or crashes).


evqna:
We already have those, we were specifically missing a test case that runs the entire app logic and expects that there should *not* be any issue to report.

MVrachev:
Okay then, it looks good to me.

Issue #118, opened 2019-01-22
Add documentation about false-positives


MVrachev:
**Is your feature request related to a problem? Please describe.**

It will be good to explain the users how to handle false-positives.

**Describe the solution you'd like**

It would be best if we have a documentation page about this.

MVrachev:
Closed by: https://github.com/vmware/precaution/pull/121

PR #121, opened 2019-01-23
Add false-positives documentation

MVrachev:
Closes: https://github.com/vmware/precaution/issues/118

I was thinking that it would be a good idea to give information for the users how to handle false positives.

On another hand that means that non secure code may be introduced into the main master branch because the pull reuqest author may use this feature incorrectly.

### DIFF
--- a/docs/false-pos.md
+++ b/docs/false-pos.md
@@ -3,7 +3,7 @@
 
 As with all automated detection tools there will be cases of false positives.
 
-The way to handle this situation is to write #nosec in your code
+If Precaution reports a failure that has been manually verified as being safe, it is possible to annotate the code with a `#nosec` comment.
 
 Here are examples:
 
--- a/docs/false-pos.md
+++ b/docs/false-pos.md
@@ -13,6 +13,10 @@ Although this line may cause a potential security issue, it will not be reported
 
 ``` self.process = subprocess.Popen('/bin/echo', shell=True)  # nosec ```
 
+
+If you want to read more about annotating the code:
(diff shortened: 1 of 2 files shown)


docs/false-pos.md
```diff
@@ -0,0 +1,34 @@
+
+# False positives and how to handle them
+
+No tool is perfect, neither is Precaution and false positives may happened.
```
evqna:
```suggestion
As with all automated detection tools there will be cases of false positives.
```

docs/false-pos.md
```diff
@@ -0,0 +1,34 @@
...
+
+The way to handle this situation is to write #nosec in your code
```
evqna:
```suggestion
If Precaution reports a failure that has been manually verified as being safe, it is possible to annotate the code with a `#nosec` comment.
```

docs/false-pos.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Python example
+
+Although this line may cause a potential security issue, it will not be reported:
+
+``` self.process = subprocess.Popen('/bin/echo', shell=True)  # nosec ```
+
+## Go example
+
+Same here, with #nosec comment you basically say: "I know what I am doing don't warn me":
```
evqna:
```suggestion
In Go the annotation has to be inside a line or block comment:
```

MVrachev:
That would be awesome think to have Eric, but still I am not sure how we would do this. :(
But still I will create an issue about this.

PR #125, opened 2019-01-27
Update links to properly show in docs html

ericwb:
The links in false-pos.md do not render properly in the html.  They don't show as links.  

See https://vmware.github.io/precaution/docs/false-pos.html


MVrachev:
LGTM

PR #131, opened 2019-02-05
Use one folder for Gosec and Bandit

MVrachev:
We tried before to use one folder - cache/ after we figure out there that we can't because Gosec needs a full GOPATH. We stuck with using two folders cache/ for Python and cache/go/src for Go files.

I and Antoine had discussed before that we can use one folder for both Bandit and Gosec but it would be cache/go/src.
I tried this and it worked. I removed all clutter code connecting with deciding which folder to use.

### DIFF
--- a/cache.js
+++ b/cache.js
@@ -8,16 +8,9 @@ const path = require('path')
  * Get the cache path for this PR branch tag
  * @param {number} repoID repository identifier
  * @param {number} prID pull request identifier
- * @param {string} app Optional: an app that calls the function. (default bandit)
- * It's needed because gosec uses GOPATH.
  */
-function getBranchPath (repoID, prID, app) {
-  app = app || 'bandit'
-  if (app === 'bandit') {
-    return path.join('cache', repoID.toString(), prID.toString())
-  } else if (app === 'gosec') {
-    return path.join('cache/go/src', repoID.toString(), prID.toString())
-  }
+function getBranchPath (repoID, prID) {
+  return path.join('cache/go/src', repoID.toString(), prID.toString())
 }
 
 /**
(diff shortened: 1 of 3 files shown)


cache.js
```diff
@@ -8,16 +8,9 @@ const path = require('path')
...
  */
-function getBranchPath (repoID, prID, app) {
-  app = app || 'bandit'
-  if (app === 'bandit') {
-    return path.join('cache', repoID.toString(), prID.toString())
-  } else if (app === 'gosec') {
-    return path.join('cache/go/src', repoID.toString(), prID.toString())
-  }
+function getBranchPath (repoID, prID) {
+  return path.join('cache/go/src', repoID.toString(), prID.toString())
```
ericwb: So it's a little odd that non-Go language files with be added into a go named cache directory.  It would add confusion.  Can we use a path without the name of `Go` in it?

MVrachev: Well, no we can't. Here are is the proof:
1) In the Golang documentation here is what is mentioned about the GOPATH environmental variable: https://golang.org/doc/code.html#GOPATH
2) I spoke with Antoine about this before https://github.com/vmware/precaution/pull/56#issuecomment-445218183
3) Still, I tried to change that. I changed the getBranchPAth function:
image

and here is the result from the logs from Heroku:
image

ericwb:
I feel this adds confusion by using a Go directory for non-Go files.

MVrachev:
You have a point but also that way we don't have to do a checking when we save. We can leave it that way. It's not a great problem.

MVrachev:
We decided that we will use two folders - one for Go (because of the limitation of Golang) "cache/go/src" and one for all other files "cache/".

PR #132, opened 2019-02-05
Add info about private repositories

MVrachev:
I tested can we use Precaution on private repositories and the great news is that we can do it even without adding even a line of code!

I think that is worth mentioning in the README.


MVrachev:
Okay, I agree. For now, I will remove this pull request.

Issue #138, opened 2019-02-13
Bug - deletion of a file


MVrachev:
**Describe the bug**

There is a bug when in a pull request there is a deletion of a file.

**To Reproduce**
1. Delete a file in your repository
2. Create a pull request
3. Precaution result will be "App error 
{"message":"Not Found","documentation_url":"https://developer.github.com/v3/repos/contents/#get-contents"} "

**Expected behavior**
Ignore the deleted files.

MVrachev:
I have tracked down from where the problem comes. I am searching for a solution.

PR #139, opened 2019-02-14
Fix for pull requests with deletions

MVrachev:
I noticed before that there is a bug when there is a pull request in which a file is deleted.

Closes: https://github.com/vmware/precaution/issues/138

I saw that the bug is caused by four reasons:
1) we didn't filter the right attribute within fileJSON in index.js
2) the status is 'removed' instead of 'deleted' when a file is deleted in a pull request.
3) when we use async filtering in index.js the code continues to the map and it crashes there with the message "endsWith function of undefined"
4) when we filter a removed file and this promise is resolved the result is undefined. We have to filter for that.

I added two unit tests for these kinds of pull requests.


index.js
```diff
@@ -71,7 +71,8 @@ async function runLinterFromPRData (pullRequests, context, headSha) {
 
     // For now only deal with one PR
     const PR = pullRequests[0]
-    const inputFiles = resolvedPRs[0]
+    // It happens sometimes that the resolverPR contains undefined members
```
ericwb:
```suggestion
    // Sometimes the resolverPR list contains undefined members
```

test/fixtures/pull_request.deletions.json
```diff
@@ -0,0 +1,16 @@
...
+    {
+      "filename": "cgi.py",
+      "status": "removed"
+    },
+    {
+      "filename": "hello.go",
+      "status": "removed"
+    }
+  ]
+}
```
ericwb: Fix the end-of-file

test/fixtures/pull_request.deletions.modif.add.json
```diff
@@ -0,0 +1,16 @@
...
+    {
+      "filename": "cgi.py",
+      "status": "added"
+    },
+    {
+      "filename": "hello.go",
+      "status": "modified"
+    }
+  ]
+}
```
ericwb: Fix the end-of-file

test/index.test.js
```diff
@@ -201,6 +203,47 @@ describe('Bandit-linter', () => {
...
+        conclusion: 'success',
+        owner: 'owner_login',
+        repo: 'repo_name',
+        completed_at: expect.any(String)
+      }))
+    })
+
+    test('handles PRs with deletions, modifications and additions', async () => {
+      mockPRContents(github, sampleDelAddModif)
+
```
ericwb: seems like an unnecessary blank line

test/index.test.js
```diff
@@ -201,6 +203,47 @@ describe('Bandit-linter', () => {
       }))
     })
 
+    test('handles PRs with only deletions', async () => {
+      mockPRContents(github, sampleOnlyDeletions)
+
```
ericwb: seems like unnecessary blank line

Issue #142, opened 2019-02-16
Changed code several times in PR, and marked success when shouldn't


ericwb:
**Describe the bug**
Something in the precaution scan is messing up when multiple commits are being added to a PR.  On the 3rd or so update of a file, Precaution all the sudden reported Success, when it should still have been fail.  

**To Reproduce**
https://github.com/ericwb/bandit/pull/2/checks

**Expected behavior**
I still expect errors since not all problems were resolved. Namely in this case, it no longer shows the errors on using requests module with verify=False (line 896)

**Screenshots**

**Additional context**

ericwb:
I believe the issue is there is a syntax error in the code:

```
browne-a02:travel-preapproval browne$ python test.py 
  File "test.py", line 193
    def Connect(host='localhost', port=443, user='root', pwd,
               ^
SyntaxError: non-default argument follows default argument
```
But Precaution shouldn't report overall status as Success for syntax errors.  So still problem in the reporting here.

MVrachev:
Until the third commit in your example you have a lot of issues - 20 to be exact: 
https://github.com/ericwb/bandit/pull/2/commits/f3b41ef88a427a336d561dff3a3fa9ed4b4387c7 the same amount of issues I found running Bandit locally.

Then when you look at the weak_crypto file from the fourth commit you don't get any errors by Precaution and the reason for that is because Bandit has a problem scanning the file "weak_crypto".

![image](https://user-images.githubusercontent.com/16246778/52952726-29225800-338e-11e9-8822-03db7f7295cf.png)

Overall our problem is that we don't get any feedback from Bandit when it has problems with parsing the AST.

MVrachev:
It looks like this is not only a problem to code scanned by Bandit but also about code scanned by Gosec.
This is my code: 
![image](https://user-images.githubusercontent.com/16246778/52953959-91266d80-3391-11e9-9399-5e77fb8d8966.png)

when I run go run I get:
![image](https://user-images.githubusercontent.com/16246778/52953973-9daac600-3391-11e9-8a53-90992930b14c.png)

You expect that Precaution will catch this but it did not.

MVrachev:
The problem is not that from more than 3 commits Precaution scan screws up but the problem is that Precaution doesn't catch problems from the ASTs used by Bandit and Gosec.

I created a new issue about it https://github.com/vmware/precaution/issues/144
and I think we can close this issue.

ericwb:
[...]

PR #145, opened 2019-02-19
Handle invalid python files with syntax errors

MVrachev:
Related to: https://github.com/vmware/precaution/issues/142

This pull request fixes an issue when scanning Python code with syntax errors. 
If there are invalid python files they will be annotated at their begging like this:

image

The error annotation is with high severity and confidence and that way the check run will fail because of this issue.


test/fixtures/python/crypto.invalid.py
```diff
@@ -0,0 +1,259 @@
+## Copyright 2019 VMware, Inc.
+## SPDX-License-Identifier: BSD-2-Clause
```
ericwb: Use a simpler example.  This is way too large.

linters/bandit.js
```diff
@@ -53,7 +53,16 @@ module.exports = class Bandit {
    * @param {Buffer} data The raw linter results data
    */
   parseResults (data) {
-    return JSON.parse(data)
+    let parsedData = JSON.parse(data)
+    if (parsedData.errors.length !== 0) {
+      let errStr = '\n'
+      for (let i = 0; i < parsedData.errors.length; ++i) {
+        errStr += 'issue ' + (i + 1) + ': ' + parsedData.errors[i].reason + ': ' +
```
ericwb: It looks odd to start a message in lower case.

test/fixtures/python/crypto.invalid.py
ericwb: We also shouldn't be including other projects code even though this was the code the bug was reporting with.

ericwb: You're also re-licensing code that is licensed as Apache 2

MVrachev:
> In your commit messages, please keep it less like a dialog and more to the point on what the PR is fixing.
> 
> s/AST three/AST
> 
> "Before we didn't check" - this sounds redundant because every bug fix could start that way.
> 
> Are we checking the exit code of the process that we spawn?

I edited the pr message.

Unfortunately, we can't use the exit code of the Bandit and Gosec processes... :( 

I ran Bandit on two invalid files and checked the exit code of the command:

image

I did the same think with Gosec:

image

test/fixtures/python/crypto.invalid.py
MVrachev: I agree. I will change that example.

linters/bandit.js
MVrachev: Well, the full message looks like the one from the picture.

test/fixtures/python/crypto.invalid.py
MVrachev: Yep, I just used another file.

linters/bandit.js
ericwb: The PR message still states "issue 1...."

runner.js
```diff
@@ -41,7 +41,6 @@ function run (linter, workingDirectory, files) {
   process.stderr.on('data', (chunk) => {
     errorLogs += chunk.toString()
   })
-
```
ericwb: Undo non-functional changes

test/fixtures/python/sum.invalid.py
```diff
@@ -0,0 +1,10 @@
+num = eval(input("Enter a positive number: "
+if num < 0:
+   print("Non positive number"
+else:
+   sum = 0
+   # use while loop to iterate un till zero
+   while num > 0):
+       sum += num
+       num -= 1
+   print("The sum is",sum)
```
ericwb: And newline

test/fixtures/python/sum.invalid.py
```diff
@@ -0,0 +1,10 @@
+num = eval(input("Enter a positive number: "
```
ericwb: The syntax errors here are largely the same as the other example.  Why do we need two examples?  I would understand having one for Python and one for Go.  But two for Python that have missing ( or ) doesn't give much value.

Also, you're missing the copyright and SPDX

runner.js
```diff
@@ -56,9 +55,13 @@ function reportHandler (linter, reportFilePath, resolve, reject, logs) {
...
       return reject(err)
     } else {
-      const results = linter.parseResults(data)
-      const report = linter.generateReport(results)
-      return resolve(report)
+      try {
+        const results = linter.parseResults(data)
+        const report = linter.generateReport(results)
+        return resolve(report)
+      } catch (err) {
```
ericwb: So when this runs, does it immediately  fail and quit the scan?  Or does it scan all files, report on what failed due to syntax errors and what failed due to security issues?

ericwb:
In your PR message 
```This issue is easily resolved when using Bandit but the situation with Gosec is a little more complicated. I will try to create a patch for this there.```

This sounds like you're casually talking to someone and not describing the fix.

```When this pr is merged``` - no need to state this as this would end up in the history and read funny.

```This pull request gives a solution when scanning invalid python code``` - just state `This fixes an issue when scanning Python code with syntax errors`

File: test/fixtures/python/sum.invalid.py (hunk 5 above)
Comment (MVrachev) [reply test/fixtures/python/sum.invalid.py]: I wanted to see how will look the output when we have two bad files.

File: runner.js (hunk 6 above)
Comment (MVrachev) [reply runner.js]: Right now it fails when there is an invalid file.
I will change this.

**Comment by MVrachev** [comment]
I closed the pull request by accident. 

I will do as you suggested. I will just create an error annotation that will show on the top of the invalid file and that way we can report the other issues we find.

File: linters/bandit.js [hunk 7]
```diff
@@ -53,7 +53,24 @@ module.exports = class Bandit {
...
+    if (parsedData.errors.length !== 0) {
+      for (let i = 0; i < parsedData.errors.length; ++i) {
+        let errAnnotation = {
+          filename: parsedData.errors[i].filename,
+          line_number: 1,
+          issue_severity: 'HIGH',
+          issue_confidence: 'HIGH',
+          issue_text: '',
+          test_id: 'ERROR',
+          test_name: 'Invalid file' }
```
ericwb: Instead of "invalid file" can't we reliably say syntax error?

linters/bandit.js
```diff
@@ -53,7 +53,24 @@ module.exports = class Bandit {
    * @param {Buffer} data The raw linter results data
    */
   parseResults (data) {
-    return JSON.parse(data)
+    let parsedData = JSON.parse(data)
+
+    if (parsedData.errors.length !== 0) {
```
ericwb: Be careful with if != checks on lengths.  Some functions may return a -1 to indicate conditions.  As a result ever time I see a check like this, alarm bells go off in my head.  :)

test/fixtures/python/sum.invalid.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2019 VMware, Inc.
+# SPDX-License-Identifier: BSD-2-Clause
+
+num = eval(input("Enter a positive number: "
+if num < 0:
+   print("Non positive number"
```
ericwb: Instead of doing the same type of syntax error (missing paren), try using an invalid keyword or something else.

linters/bandit.js
MVrachev: Thank you, that's good advice and a reminder.

linters/bandit.js
MVrachev: Yes, that naming makes more sense but are you sure that only syntax errors can be the reason for an invalid python file?

test/fixtures/python/sum.invalid.py
MVrachev: Yep, I will.

linters/bandit.js
```diff
@@ -53,7 +53,24 @@ module.exports = class Bandit {
...
+    if (parsedData.errors.length > 0) {
+      for (let i = 0; i < parsedData.errors.length; ++i) {
+        let errAnnotation = {
+          filename: parsedData.errors[i].filename,
+          line_number: 1,
+          issue_severity: 'HIGH',
+          issue_confidence: 'HIGH',
+          issue_text: '',
+          test_id: 'ERROR',
+          test_name: 'Syntax errors' }
```
ericwb: errors is plural indicating more than one.  we don't know for sure there are multiple syntax errors.

test/bandit.spawn.test.js
```diff
@@ -33,6 +33,17 @@ describe('Bandit runner', () => {
...
 
+  test('Handles when there are invalid and valid python files', async () => {
+    let mixedReport = await bandit('test/fixtures/python', ['sum.invalid.py', 'https.py'])
+
+    expect(mixedReport.annotations.length).toBe(4)
+    expect(mixedReport.annotations[0].title).toBe('B309:blacklist')
+    expect(mixedReport.annotations[3].path).toEqual('sum.invalid.py')
+    expect(mixedReport.annotations[3].start_line).toBe(1)
+    expect(mixedReport.annotations[3].annotation_level).toBe('failure')
+    expect(mixedReport.annotations[3].title).toBe('ERROR:Syntax error')
```
ericwb: How come this doesn't fail? I'd expect `Error: Syntax error` not `ERROR:Syntax error`

MVrachev:
Sadly, if the start line and end line in one annotation are 0 then the annotation won't show up in the "Files Changed" tab.

I tried to use start line 0 and end line 1 but the result was that the annotation showed like a mistake on line 1:
image

So yes, we are stuck with the "Error syntax" annotations to be shown on line 1...

test/bandit.spawn.test.js
MVrachev: Well, right now we do isssue_id:issue_name and there is no space. I sent a pull request where I add spaces.

bandit/bandit_report.js
```diff
@@ -12,12 +12,15 @@ function createMoreInfoLinks (issues) {
...
 
   for (let issue of issues) {
-    if (issuesMap.has(issue.test_id) === false) {
+    if (issuesMap.has(issue.test_id) === false && issue.more_info) {
       issuesMap.set(issue.test_id)
       const text = `${issue.test_id}: ${issue.test_name}`
       moreInfo += `[${text}](${issue.more_info})\n`
     }
   }
+  if (issuesMap.size <= 0) {
```
ericwb: curious, does javascript ever let a size be less than 0?

linters/bandit.js
```diff
@@ -53,7 +53,22 @@ module.exports = class Bandit {
...
+
+    for (let error of parsedData.errors) {
+      let errAnnotation = {
+        filename: error.filename,
+        line_number: 1,
+        issue_severity: 'HIGH',
+        issue_confidence: 'HIGH',
+        issue_text: '',
+        test_id: 'ERROR',
+        test_name: 'Syntax error' }
```
ericwb: Put trailing } on next line to be consistent style with all the other code.

bandit/bandit_report.js
MVrachev: I read a little more about this and all answers says that it can't be negative BUT there is one curious case:
image

mentioned here: https://stackoverflow.com/questions/32386253/can-the-js-array-length-property-ever-return-a-negative-value

linters/bandit.js
MVrachev: Yes, you are right. I wonder how do you notice those things. xD

PR #147, opened 2019-02-21
Fix Gosec annotation

MVrachev:
Closes: #148 

When we refactored our code it looks like we forgot to give the right path to the issues found by Gosec. The result is that the issues are shown in the checks page but not into the "Files changed" page under the code problematic lines.

The problem is that Gosec gives the absolute path to the found problem but we are interested only to the relative path against the current working directory.

### DIFF
--- a/runner.js
+++ b/runner.js
@@ -33,6 +33,7 @@ function run (linter, workingDirectory, files) {
   const filtered = linter.filter(files)
 
   if (filtered.length === 0) { return linter.defaultReport }
+  
   const reportFilePath = path.join(workingDirectory, '..', linter.reportFile)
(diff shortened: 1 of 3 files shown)


runner.js
```diff
@@ -33,7 +33,6 @@ function run (linter, workingDirectory, files) {
   const filtered = linter.filter(files)
 
   if (filtered.length === 0) { return linter.defaultReport }
-
```
ericwb: non-functional change. please undo

runner.js
```diff
@@ -33,7 +33,7 @@ function run (linter, workingDirectory, files) {
...
+
```
ericwb: accidental space added

PR #149, opened 2019-02-28
Update Gosec version

MVrachev:
Gosec updated its version. Now Gosec supports reports for Golang errors which will be really useful for us to report to the user.


MVrachev:
I will create a separate pull request adding Golang error annotations and this.

PR #154, opened 2019-03-01
Handle invalid go files with syntax errors

MVrachev:
Related to: https://github.com/vmware/precaution/issues/142

From version 1.3 of Gosec, we get information about the possible syntax errors found from the golang parser.
That's something important to report to the user because if there are syntax errors in a file the file won't be scanned for security vulnerabilities.

Additionally, to reporting the golang syntax errors I added a unit test about this case.

### DIFF
--- a/gosec/gosec_annotations.js
+++ b/gosec/gosec_annotations.js
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 
 const { getAnnotationLevel } = require('../annotations_levels')
+const { config } = require('../config')
 
 /**
  * @param {*} issue - an issue from which the annotation will be build
@@ -15,7 +16,8 @@ function getAnnotation (issue, directory) {
   const endLine = startLine
   const annotationLevel = getAnnotationLevel(issue.severity, issue.confidence)
   const title = `${issue.rule_id}: ${issue.details}`
-  const message = `The issue is in the code: ${issue.code}`
+  const message = title === config.syntaxErrorTitle ? `${issue.code}`
+    : `The issue is in the code: ${issue.code}`
 
   return {
     path,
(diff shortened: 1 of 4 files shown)


gosec/gosec_annotations.js
```diff
@@ -15,7 +16,8 @@ function getAnnotation (issue, directory) {
   const endLine = startLine
   const annotationLevel = getAnnotationLevel(issue.severity, issue.confidence)
   const title = `${issue.rule_id}: ${issue.details}`
-  const message = `The issue is in the code: ${issue.code}`
+  const message = title === config.syntaxErrorTitle ? `${issue.code}`
+    : `The issue is in the code: ${issue.code}`
```
MVrachev: I added this check because it did make sense to me to report a syntax issue like:
image

and still was really useful to report the security issues like that:
image

config.js
```diff
@@ -7,7 +7,8 @@ const config = {
   checkRunName: 'Precaution',
   noIssuesResultTitle: 'No issues found',
   noIssuesResultSummary: 'There were no issues found.',
-  issuesFoundResultTitle: 'Issues found'
+  issuesFoundResultTitle: 'Issues found',
```
ericwb: Same change is in PR #145.  It's best to just create a commit with a dependency.

gosec/gosec_annotations.js
ericwb: `The issue is in the code` sounds awkward and obvious.  In your second example, what is the issue?

MVrachev: I used the prefix: `The issue is in the code` to annotate where exactly in the line is the problem. Bandit gives you information about the exact code causing the problem so do Gosec and I wanted to use that.

config.js
MVrachev: I am not sure how to do that.

MVrachev: I will now.

gosec/gosec_annotations.js
ericwb: I say that because we already know it's an issue and we already know its source code.  So sounds awkward in that way.

MVrachev: Okay, then I can remove that.

PR #156, opened 2019-03-08
Update jest to fix a bug in Travis CI

MVrachev:
There is a bug when Travis CI uses jest. This bug is fixed in the latest versions of jest and that's why we need to upgrade.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "probot": "^7.4.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


ericwb:
You should either describe the bug or reference it.  Reading the PR message, I have no idea what this fixes.

MVrachev:
Here is the bug: https://github.com/facebook/jest/issues/8070

Issue #134, opened 2019-02-06
Handle pull requests with over 30 diff files


MVrachev:
**Describe the bug**

There is a paging limit in the GitHub API for how many files we can list/download:
https://developer.github.com/v3/pulls/#list-pull-requests-files

But we can change that limit up to 100 (I think): 
https://developer.github.com/v3/#rate-limiting

The question is how we can do that using Javascript not curl bash command? 

**Expected behavior**

A solution when we will be able to download more than 30 files or at least up to 100 with one query if possible. If we can't download them with one query then maybe with multiple queries.

MVrachev:
I created a pull request with 70 files and the result was this:
![image](https://user-images.githubusercontent.com/16246778/53426135-15f73400-39ef-11e9-95e9-1cebd07c84d9.png)

MVrachev:
Fixed by https://github.com/vmware/precaution/pull/157 and https://github.com/vmware/precaution/pull/158

PR #157, opened 2019-03-08
Download more than 30 files

MVrachev:
Related to: https://github.com/vmware/precaution/issues/134

In order to download all of the diff files from a pull request, we have to access the data from GitHub page by page.

Also, there is an option to customize how many results do you want per page.

I added a unit test handling the paging but I  had to change the other unit tests a little bit in order to achieve that.

### DIFF
--- a/config.js
+++ b/config.js
@@ -8,7 +8,8 @@ const config = {
   noIssuesResultTitle: 'No issues found',
   noIssuesResultSummary: 'There were no issues found.',
   issuesFoundResultTitle: 'Issues found',
-  syntaxErrorTitle: 'ERROR: Syntax error'
+  syntaxErrorTitle: 'ERROR: Syntax error',
+  amountFilesListedPerPage: 30
 }
 
 const annotationsLevels = {
--- a/config.js
+++ b/config.js
@@ -9,7 +9,7 @@ const config = {
   noIssuesResultSummary: 'There were no issues found.',
   issuesFoundResultTitle: 'Issues found',
   syntaxErrorTitle: 'ERROR: Syntax error',
-  amountFilesListedPerPage: 30
+  numberFilesListedPerPage: 30
 }
 
 const annotationsLevels = {
--- a/config.js
+++ b/config.js
@@ -9,7 +9,7 @@ const config = {
   noIssuesResultSummary: 'There were no issues found.',
   issuesFoundResultTitle: 'Issues found',
   syntaxErrorTitle: 'ERROR: Syntax error',
-  numberFilesListedPerPage: 30
+  numFilesPerPage: 30
 }
 
 const annotationsLevels = {
--- a/github_api_helper.js
+++ b/github_api_helper.js
@@ -26,6 +26,14 @@ function inProgressAPIresponse (context, headSha) {
   })
 }
 
+// Filters the files which are not relevant to us
+function filterData (rawData) {
+  return rawData.data.filter(file => config.fileExtensions.reduce((acc, ext) => acc || file.filename.endsWith(ext), false))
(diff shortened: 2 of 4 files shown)


test/index.test.js
```diff
@@ -19,7 +20,11 @@ const sampleOnlyDeletions = require('./fixtures/pull_request.deletions.json')
 const sampleDelAddModif = require('./fixtures/pull_request.deletions.modif.add.json')
 
 function mockPRContents (github, PR) {
-  github.pullRequests.listFiles = jest.fn().mockResolvedValue(PR)
+  github.pullRequests.listFiles = jest.fn(() => {
+    --github.numPagesOfContent
+    return PR
```
ericwb: We shouldn't use variables in all caps.  It effectively looks like a const

test/index.test.js
```diff
@@ -241,6 +259,25 @@ describe('Bandit-linter', () => {
...
     })
 
+    test('handles paging through the GitHub API', async () => {
+      github.numPagesOfContent = 2
+      await app.receive(pullRequestOpenedEvent)
+
+      expect(github.hasNextPage).toHaveBeenCalledTimes(2)
+      expect(github.getNextPage).toHaveBeenCalledTimes(1)
+
+      const expectedRes = { 'data':
```
ericwb: put 'data': on the next line

github_api_helper.js
```diff
@@ -26,21 +26,37 @@ function inProgressAPIresponse (context, headSha) {
   })
 }
 
+// Filters the files which are not relevant to us
```
ericwb: Can you please add a proper function doc

github_api_helper.js
```diff
@@ -26,21 +26,37 @@ function inProgressAPIresponse (context, headSha) {
...
+    .filter(fileJSON => fileJSON.status !== 'removed').map(async fileInfo => {
+      return fileInfo.filename
+    })
+}
+
 /**
  * Get list of files modified by a pull request
  * @param {import('probot').Context} context Probot context
  * @param {number} number the pull request number inside the repository
  * @returns {Promise<any>} GitHub response
```
ericwb: Doesn't look like you're returning a GitHub response object any longer.

MVrachev: Yes, a good catch.

index.js
```diff
@@ -102,25 +102,19 @@ async function processPullRequest (pullRequest, context) {
...
+  const filesDownloadedPromise = response
+    .map(async fileName => {
+      const headRevision = apiHelper.getContents(context, fileName, ref, pullRequest.head)
 
       // TODO: merge this code with linter-specific path resolution
-      if (filename.endsWith('.py')) {
-        cache.saveFile(repoID, id, filename, (await headRevision).data, 'python')
-      } else if (filename.endsWith('.go')) {
-        cache.saveFile(repoID, id, filename, (await headRevision).data, 'go')
+      if (fileName.endsWith('.py')) {
```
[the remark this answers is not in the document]
MVrachev: Okay, but I don't see why we can't change filename => fileName now. :D

test/index.test.js
MVrachev: It used to be with big letters before but you are right.

config.js
```diff
@@ -8,7 +8,8 @@ const config = {
   noIssuesResultTitle: 'No issues found',
   noIssuesResultSummary: 'There were no issues found.',
   issuesFoundResultTitle: 'Issues found',
-  syntaxErrorTitle: 'ERROR: Syntax error'
+  syntaxErrorTitle: 'ERROR: Syntax error',
+  numberFilesListedPerPage: 30
```
ericwb:
```suggestion
  numFilesPerPage: 30
```

MVrachev: I will change it.

PR #158, opened 2019-03-08
Send more than 50 issues to GitHub

MVrachev:
Related to: https://github.com/vmware/precaution/issues/134

GitHub has a limitation on how many annotations can you send with one single API call.

The solution of that problem, as pointed out in the documentation is to make more API calls until you send all of your annotations

I added a unit test for that case.


github_api_helper.js
```diff
@@ -98,22 +98,48 @@ function getConclusion (annotations) {
  * @param {import('probot').Context} context Probot context
  * @param {Number} runID check run identifier
  * @param {Object} output merged scan output
+ * @param {Number} annotationsPerPage Optional: number of annotations to send with one API call.
```
ericwb: How is this optional?  Will it work if I don't pass a value?  I don't see a default value

github_api_helper.js
```diff
@@ -98,22 +98,48 @@ function getConclusion (annotations) {
...
-  return context.github.checks.update({
-    check_run_id: runID,
-    owner,
-    repo,
-    status: 'completed',
-    completed_at: completedAt,
-    conclusion: getConclusion(output.annotations),
-    output
-  })
+  let timesToSendUpdates = Math.ceil(output.annotations.length / numAnnotationsLeftToSend)
```
ericwb: variable name is a little odd.  I would think this is "times" as a date/time

github_api_helper.js
```diff
@@ -98,22 +98,48 @@ function getConclusion (annotations) {
...
  * @returns {Promise<any>} GitHub response
  * See: https://developer.github.com/v3/checks/runs/#update-a-check-run
  */
-function sendResults (context, runID, output) {
+function sendResults (context, runID, output, annotationsPerPage) {
   const { owner, repo } = context.repo()
-  const completedAt = new Date().toISOString()
+  let numAnnotationsLeftToSend = annotationsPerPage || output.annotations.length
+  if (numAnnotationsLeftToSend <= 0 || numAnnotationsLeftToSend > 50) {
+    numAnnotationsLeftToSend = 50
```
ericwb: This arbitrary number of `50` here could use a constant instead.  You use it multiple times too.

github_api_helper.js
MVrachev: I refactored the function because it was unclear what I was doing while using those variable names.
I must have been tired. :D

github_api_helper.js
MVrachev: Yep, I will change that.

github_api_helper.js
MVrachev: There is a default value but it was not clear.

ericwb: Any plans to change this?

github_api_helper.js
```diff
@@ -118,22 +118,50 @@ function getConclusion (annotations) {
...
-    check_run_id: runID,
-    owner,
-    repo,
-    status: 'completed',
-    completed_at: completedAt,
-    conclusion: getConclusion(output.annotations),
-    output
-  })
+  let numAnnotationsPerAPICall = annotationsPerPage || config.numAnnotationsPerUpdate
+  if (numAnnotationsPerAPICall <= 0 || numAnnotationsPerAPICall > 50) {
```
ericwb: 50 is used several times, making it deserving of a constant

github_api_helper.js
MVrachev: Yes, I just changed that.

github_api_helper.js
MVrachev: Yep, I added a constant.

github_api_helper.js
MVrachev: I want to keep that parameter because I use it in my unit test for multiple annotations.
There as you can see: 
image

I want to send 2 annotations per API call which means that I will call two times github.checks.update which is want I want to test.

github_api_helper.js
```diff
@@ -118,22 +118,51 @@ function getConclusion (annotations) {
...
+ * @param {Number} annotationsPerPage Optional: number of annotations to send with one API call.
  * @returns {Promise<any>} GitHub response
  * See: https://developer.github.com/v3/checks/runs/#update-a-check-run
  */
-function sendResults (context, runID, output) {
+function sendResults (context, runID, output, annotationsPerPage) {
   const { owner, repo } = context.repo()
-  const completedAt = new Date().toISOString()
+  let numAnnotationsLeftToSend = output.annotations.length
+  const MAX_ANNOTATIONS_CALL = 50
```
ericwb: config.js also defines a 50 const.  Can we use only one definition of 50?

MVrachev: Yes, but I thought that one day we can decide to change the const numAnnotationsPerUpdate in the config.js and then we need to have the const MAX_ANNOTAIONS_CALL for the check.

ericwb: Ah I see.

Issue #160, opened 2019-03-22
Add TSLint plugin into Precaution


MVrachev:
**Is your feature request related to a problem? Please describe.**

I just found out that there is a way to make [TSLint](https://github.com/palantir/tslint) to support both TypeScript and JavaScript.

That means that we can use [tsling-config-security](https://github.com/webschik/tslint-config-security) plugin to scan both JavaScript and TypeScript.

**Describe the solution you'd like**

I will add two pull requests:

- [x] TSLint configuration
- [x] TSLint report generation
- [x] Refactor code for linter besides Bandit and Gosec
- [x] TSLint spawn
- [x] Update some unit tests to include js and ts files
- [x] Update the documentation and architecture picture with TSLint

MVrachev:
We only need to update docs/index.md for the support of TypeScript and JavaScript.

PR #163, opened 2019-03-22
Add TSLint security plugin report files and tests

MVrachev:
Part of the work outlined in the enhancement issue: https://github.com/vmware/precaution/issues/160

I added the report generation from the TSLint plugin output.

I added three unit tests.


tslint/tslint_annotations.js
```diff
@@ -0,0 +1,45 @@
...
+// SPDX-License-Identifier: BSD-2-Clause
+
+const { getAnnotationLevel } = require('../annotations_levels')
+
+/**
+ * Translates the tslint severity from "ERROR" | "WARNING" | "OFF"
+ * to the "HIGH", "MEDIUM" and "LOW"
+ * @param {String} severity issues severity given by tslint
+ */
+function translateSeverity (severity) {
```
ericwb: Aren't we doing similar mappings like this in annotations_levels.js?

tslint/tslint_annotations.js
```diff
@@ -0,0 +1,45 @@
...
+  switch (severity) {
+    case 'ERROR': result = 'HIGH'; break
+    case 'WARNING': result = 'MEDIUM'; break
+    case 'OFF': result = 'LOW'
+  }
+  return result
+}
+
+/**
+ * Create bandit Annotations
```
ericwb: bandit?

tslint/tslint_annotations.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param {*} issue
+ * @param {String} directory working directory for the Gosec process
```
ericwb: Gosec?

tslint/tslint_annotations.js
MVrachev: Yes, a mistake.

tslint/tslint_annotations.js
MVrachev: Same here.

tslint/tslint_annotations.js
MVrachev: We don't have to.
The severity attribute in the output from Bandit and Gosec is "HIGH", "MEDIUM" or "LOW" like the ones used in the annotations_levels.js.
This is a specific situation in this linter.

But there is another annoying thing. There is no "confidence" in the output and I give "HIGH" confidence by default...

ericwb: So yes, `annotations_levels.js` looks at confidence whereas TSLint has none, but that's a very easy change to use a `default` in the switch statement for cases where there is no confidence.

Also, looks odd to map `OFF` to `LOW` since I assume `OFF` indicates you'll get no results.  Although I'm not clear what happens with `OFF` according to the documentation.

MVrachev: So we should ignore the issues with annotations off?

I will speak with the TSLint team about it.

MVrachev: When there is no confidence from a linters output what should be the default value? HIGH?

tslint/tslint_annotations.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * Create TSLint Annotations
+ * @param {*} issue
+ * @returns {Object} see https://developer.github.com/v3/checks/runs/#annotations-object-1
+ */
+function getAnnotation (issue) {
+  const path = issue.name
+  // I add 1 to startLine and endline because they are off by 1 as described here:
```
ericwb: Who is I?  Try to avoid using first person in comments

tslint/tslint_annotations.js
```diff
@@ -0,0 +1,46 @@
...
+ * Translates the tslint severity from "ERROR" | "WARNING" | "OFF"
+ * to the "HIGH", "MEDIUM" and "LOW"
+ * @param {String} severity issues severity given by tslint
+ */
+function translateSeverity (severity) {
+  let result = ''
+  switch (severity) {
+    case 'ERROR': result = 'HIGH'; break
+    case 'WARNING': result = 'MEDIUM'; break
+    case 'OFF': result = 'LOW'
```
ericwb: It's still not clear to me what OFF severity means.

annotations_levels.js
```diff
@@ -38,20 +38,23 @@ function getAnnotationLevel (severity, confidence) {
         case 'HIGH' : result = annotationsLevels.severityHIGHconfidenceHIGH; break
         case 'MEDIUM' : result = annotationsLevels.severityHIGHconfidenceMEDIUM; break
         case 'LOW' : result = annotationsLevels.severityHIGHconfidenceLOW; break
+        default: result = annotationsLevels.severityHIGHconfidenceHIGH
```
ericwb: Since we're not using these changes for this PR, I would ask you put this in a separate PR.

tslint/tslint_annotations.js
```diff
@@ -0,0 +1,46 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: BSD-2-Clause
+
+const { getAnnotationLevel } = require('../annotations_levels')
+
+/**
+ * Translates the tslint severity from "ERROR" | "WARNING" | "OFF"
+ * to the "HIGH", "MEDIUM" and "LOW"
+ * @param {String} severity issues severity given by tslint
```
ericwb: What is meant by "severity issues severity"

tslint/tslint_annotations.js
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Create TSLint Annotations
+ * @param {*} issue
+ * @returns {Object} see https://developer.github.com/v3/checks/runs/#annotations-object-1
+ */
+function getAnnotation (issue) {
+  const path = issue.name
+  // I add 1 to startLine and endline because they are off by 1 as described here:
+  // https://github.com/palantir/tslint/issues/4528
```
ericwb: I also generally don't add comments referring to bugs since that is what git log and git blame is for.

annotations_levels.js
MVrachev: I just realized that we are giving a default annotation level when we initialize the result value in the begging of the function. The default annotation level when there is no confidence provided is "warning" whatever the severity is.

Do we want to give other default annotation level based on the severity when there is no confidence provided?

MVrachev: I will remove the default statements from this switch for now.

tslint/tslint_annotations.js
MVrachev: I will paraphrase it.

tslint/tslint_annotations.js
MVrachev: To we fair I am not sure after spending a few hours investigating. 
I opened an issue about it: https://github.com/palantir/tslint/issues/4634

tslint/tslint_annotations.js
MVrachev: Yep, you are right. I will change it.

tslint/tslint_report.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: BSD-2-Clause
+
+const { getAnnotation } = require('./tslint_annotations')
+const { countIssueLevels } = require('../annotations_levels')
+
+/**
+ * Process TSLint output (generate annotations, count issue levels)
+ * @param {*} results TSLint json output
+ */
```
ericwb: Missing doc on return

tslint/tslint_annotations.js
```diff
@@ -0,0 +1,46 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: BSD-2-Clause
+
+const { getAnnotationLevel } = require('../annotations_levels')
+
+/**
+ * Translates the tslint severity from "ERROR" | "WARNING" | "OFF"
+ * to the "HIGH", "MEDIUM" and "LOW"
+ * @param {String} severity the severity of the issue provided by  TSLint
+ */
```
ericwb: Missing doc about the return value

PR #164, opened 2019-03-22
Add configuration for TSLint

MVrachev:
Part of the work outlined in the enhancement issue: https://github.com/vmware/precaution/issues/160

I added configuration for TSLint to scan both JavaScript and TypeScript.


tslint/tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+  "compilerOptions": {
+    "strict": true,
+    "allowJs": true
+  },
+  "include": [
+    "./**/*"
+  ],
+  "exclude":[
+    "node_modules",
+    "coverage"
```
ericwb: What about test directories?

MVrachev: Yes, you are right but I won't need these options at all. These options will be useful if you scan your project locally and use --project flag to scan the whole project and with those options, you exclude folders you don't want to be tested.

In our case, we will have the exact paths to the files and we won't have to use which files to include or exclude.
I will remove these attributes as a whole in the next commit.

PR #171, opened 2019-04-05
Update a test comment

MVrachev:
I change the "Bandit linter" for this group of tests because it's not relevant with what they are testing.
Those tests are truly for the Precaution workflow and it's interaction with GitHub.

### DIFF
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -27,7 +27,7 @@ function mockPRContents (github, pr) {
   github.getNextPage = github.pullRequests.listFiles
 }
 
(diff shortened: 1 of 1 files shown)


ericwb:
Please rephrase the commit message.  `We forgot to change the name of those group of test before.` 

Who is we? Just state what is wrong in the code and what this change will do to fix it.

MVrachev:
> Please rephrase the commit message. `We forgot to change the name of those group of test before.`
> 
> Who is we? Just state what is wrong in the code and what this change will do to fix it.

I rephrase it.

Issue #170, opened 2019-04-03
Bug: request timeout on big pull request


MVrachev:
**Describe the bug**

When you run Precaution on truly big pull requests like ones with 6000 files Precaution hangs for a while and then it's killed by Heroku.

![april1_error](https://user-images.githubusercontent.com/16246778/55488941-a6054a80-5639-11e9-8b05-0b330c353ec0.png)

**To Reproduce**
1. Create a really big pull request.
2. Look at the logs and eventually, you will see a mistake like the one above.

**Expected behavior**
Precaution should be able to handle pull requests even with that size so that's a problem.

**Possible reason**
I believe that the problems come from the fact that Heroku will kill our web dyno (application) if a request hangs for more than 30 seconds: https://devcenter.heroku.com/articles/request-timeout

If that is truly the case then our app is little slow and is doing some heavy lifting stuff in a synchronous way and that's why it is slow.

PR #172, opened 2019-04-05
Fix for request timeout from Heroku

MVrachev:
Fixes: https://github.com/vmware/precaution/issues/170

Instead of awaiting the finish of the runLinterFromPRData function it's required by probot to execute it asynchronously so probot can send a response for the inial request.

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,14 +23,14 @@ module.exports = app => {
       // TODO: investigate GitHub editor not triggering check suite
       if (pullRequests.length > 0) {
         // Commit was pushed to a PR branch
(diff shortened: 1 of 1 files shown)


ericwb:
I think your commit message understates the problem.  It's not just "better" that the code executes asynchronously, it's required for probot to properly send a response for the initial check_run request.

MVrachev:
> I think your commit message understates the problem. It's not just "better" that the code executes asynchronously, it's required for probot to properly send a response for the initial check_run request.

I changed the message of the pr.

Issue #103, opened 2019-01-15
Clean up usage of raw strings


evqna:
The strings 'bandit' and 'gosec' appear in several places in the code, it may be time to clean that up.

MVrachev:
How do you imagine this cleanup? 
You will put a const variable called 'bandit' or 'python_linter' or what?

evqna:
Preferably using an enumeration or some other type-safe construct that prevents mis-typing or introducing bugs by not keeping the different uses in sync

evqna:
A gloal const string could also work but it still requires a bunch of code switches in a few places to call the correct code for each linter.

PR #173, opened 2019-04-05
Refactor for new linters

MVrachev:
Connected with: https://github.com/vmware/precaution/issues/160

Refactor code for new linters In our code on many places is like there will be no other linters in the future besides Bandit and Gosec.
That is something worth to change.

### DIFF
--- a/cache.js
+++ b/cache.js
@@ -1,4 +1,4 @@
-// Copyright 2018 VMware, Inc.
+// Copyright 2019 VMware, Inc.
 // SPDX-License-Identifier: BSD-2-Clause
 
 const fs = require('fs-extra')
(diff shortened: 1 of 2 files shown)


cache.js
```diff
@@ -42,7 +41,7 @@ function saveFileToPRCache (repoID, prID, filePath, data, fileType) {
  * @param {number} prID pull request identifier
  */
 function clearPRCache (repoID, prID) {
-  fs.removeSync(getBranchPath(repoID, prID, 'bandit'))
+  fs.removeSync(getBranchPath(repoID, prID, 'other linters'))
```
joshuagl: Elsewhere in the code we say `'other linter'` not `'other linters'`, might be worth defining a constant?

MVrachev: That's not a bad idea.
It's connected with this issue: 
https://github.com/vmware/precaution/issues/103

This is an optional parameter. Maybe I can just remove it?
I place it there to be more explicit. What do you think?

joshuagl: Personally I'd remove it, the function documentation lists the parameter as optional.
Btw the docstring needs updating, it says:
`@param {string} app Optional: an app that calls the function. (default bandit)`
but the default has now changed

MVrachev: Maybe a better way to do this is by making this argument as a flag.
We need that argument because of how Golang handles paths and that it has to use GOPATH.

Maybe a good idea will be to create this as a boolean flag.
If it's up it can mean "that's a golan files" or "gosec".

joshuagl: I didn't mean to suggest we remove the parameter from the function entirely. Rather instead we could explicitly set the default parameter using JavaScript's [default function parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/default_parameters) support and then only pass the parameter when we need the non-default.

A boolean flag wouldn't be the best idea, because we'd then be exposing backend specific functionality in a generic function.

joshuagl:
Precaution is currently broken in the master branch when scanning JS/TS repos. cache.js:`getBranchPath()` in master only supports *bandit* and *gosec*.

Would be good to get this PR squashed and merged.

PR #174, opened 2019-04-05
TSLint spawn functionality

MVrachev:
I added a TSLint spawn functionality + unit tests for that.

Because the TSLint rules for TypeScript are 100% compatible with JavaScript (those are the words from the maintainer of the TSLint security rules project) I tested only on vulnerable JavaScript file.

Sadly, TSLint doesn't report you when you try to scan an invalid file with syntax errors. That's why I haven't added more sophisticated functionality in parseResults function in linters/tslint.js 

### DIFF
--- a/linters/index.js
+++ b/linters/index.js
@@ -3,8 +3,10 @@
 
 const Bandit = require('./bandit')
 const Gosec = require('./gosec')
+const TSLint = require('./tslint')
 
 module.exports = {
   BANDIT: new Bandit(),
-  GOSEC: new Gosec()
+  GOSEC: new Gosec(),
+  TSLint: new TSLint()
 }
--- a/linters/index.js
+++ b/linters/index.js
@@ -8,5 +8,5 @@ const TSLint = require('./tslint')
 module.exports = {
   BANDIT: new Bandit(),
   GOSEC: new Gosec(),
-  TSLint: new TSLint()
+  TSLINT: new TSLint()
 }
--- a/linters/tslint.js
+++ b/linters/tslint.js
@@ -1,0 +1,68 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: BSD-2-Clause
+
+const cache = require('../cache')
+const report = require('../tslint/tslint_report')
+
+module.exports = class TSLint {
+  get name () {
+    return 'tslint'
+  }
+
+  /**
+   * The name of the generated report file
+   */
+  get reportFile () {
+    return 'tslint.json'
+  }
+
+  get defaultReport () {
+    return { annotations: [], issueCount: { errors: 0, warnings: 0, notices: 0 }, moreInfo: '' }
+  }
+
+  /**
+   * Retains files that can be analyzed by this linter
+   * @param {string[]} files Names of files to analyze
+   * @returns {string[]} Filtered list of file names
+   */
+  filter (files) {
+    return files.filter(name => name.endsWith('.js') || name.endsWith('ts'))
+  }
+
+  /**
+   * Returns the working directory for this analysis
+   * @param {string} repoID Unique repository id
+   * @param {string} prID PR id in repository
+   */
+  workingDirectoryForPR (repoID, prID) {
+    return cache.getBranchPath(repoID, prID, 'tslint')
+  }
+
+  /**
+   * Builds the command line args to pass to the linter process
+   * @param {string[]} files List of files to analyze
+   * @param {string} reportPath Path to the report file relative to working directory
+   */
+  args (files, reportPath) {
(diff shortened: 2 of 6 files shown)


linters/index.js
```diff
@@ -3,8 +3,10 @@
 
 const Bandit = require('./bandit')
 const Gosec = require('./gosec')
+const TSLint = require('./tslint')
 
 module.exports = {
   BANDIT: new Bandit(),
-  GOSEC: new Gosec()
+  GOSEC: new Gosec(),
+  TSLint: new TSLint()
```
joshuagl: Is there an implied convention here that all exports are upper case? We don't seem to have an explicit coding style in [CONTRIBUTING.md](https://github.com/vmware/precaution/blob/026b54e77efe9d8f2a50356fdc6ee37d11b93722/CONTRIBUTING.md#code-style) yet, but this would match [npm-coding-style](https://docs.npmjs.com/misc/coding-style.html#case-naming-etc), at least.

linters/tslint.js
```diff
@@ -0,0 +1,68 @@
...
+    return cache.getBranchPath(repoID, prID, 'tslint')
+  }
+
+  /**
+   * Builds the command line args to pass to the linter process
+   * @param {string[]} files List of files to analyze
+   * @param {string} reportPath Path to the report file relative to working directory
+   */
+  args (files, reportPath) {
+    return ['-c', '../../../tslint/tslint.json', '--format', 'json', '-o', reportPath, ...files]
```
joshuagl: What's the 2nd array item here (`'../../../tslint/tslint.json'`), it looks like it would be similar to reportPath?

test/fixtures/typescript/vulnerable.js
```diff
@@ -0,0 +1,28 @@
+const fs = require('fs-extra')
+var stdin = process.openStdin()
+
+function bildngSQLQuery () {
```
joshuagl: Is there a typo in this function name?

test/fixtures/typescript/vulnerable.js
```diff
@@ -0,0 +1,28 @@
...
+  const userId = 1
+  const query = `SELECT * FROM users WHERE id = ${userId}`
+  return query
+}
+
+function getTemplateGreeting (name) {
+  let classicString = `Hello ${name}`
+  return eval('`' + classicString + '`') // eslint-disable-line
```
joshuagl: Is the eslint comment still useful?

test/fixtures/typescript/vulnerable.js
joshuagl: I see this is addressed in a later commit, sorry.

test/fixtures/typescript/vulnerable.js
MVrachev: It's useful because we are using [JavaScript standard](https://standardjs.com/) and [Jest]https://jestjs.io/ to enforce style guidelines.
Before we run our unit tests JavaScript standard runs through jest to make sure that the standard JavaScript style is followed. 

In this case, JavaScript Standart reports the use of eval function as a problem and as a result, the Travis CI build will fail.
in order to escape that I use "eslint-disable-line" comment because under the hood JavaScript standard uses ESlint.
As described here: https://standardjs.com/#how-do-i-hide-a-certain-warning

linters/tslint.js
MVrachev: The second argument is the path to the configuration options for TSLint.
Right now we download and process all of the files in the directory
```
cache/<repository_id>/<pull_request_id>
```
and the configuration options for TSLint are in the folder:
```
tslint/tslint.json
```

linters/index.js
MVrachev: We are using JavaScipt style guidelines called - JavaScript Standart.
Maybe won't be a bad idea to change all exports with upper cases.

test/fixtures/typescript/vulnerable.js
joshuagl: Thanks for explaining.

linters/tslint.js
joshuagl: Ah, it's just coincidence that they are very similar? IMHO it would be useful if we didn't use similar filenames/paths for the input/output files. Perhaps we can change `reportFile()` to return _tslint-report.json_ ? We could unify on that filename pattern throughout Precaution? I'm happy for that to be handled separately.

linters/index.js
joshuagl: Either way, we should be consistent.

linters/tslint.js
MVrachev: That's the first time when we had a coincidence between the output file name and a configuration file.
I will rename the output file to "tslint_output.json".

You know when you brought this up maybe it's a good idea to change the name of the report files for the other linters too.

PR #176, opened 2019-04-05
Enable Mergify for automatic merges

ericwb:
The Mergify GitHub App is already installed to the Precaution
repo.  This commit enables the app by setting necessary configuration
via the YAML file.

Note that this config sets the following conditions
- The number of approved code reviews must be 1 or greater
- The Travis CI must pass
- Enabled on branch master only
- Don't merge PRs with work-in-progress label
- And finally, only merge with label ready-to-merge

https://doc.mergify.io

Signed-off-by: Eric Brown <browne@vmware.com>

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,12 @@
+pull_request_rules:
+  - name: merge when CI passes and 1 approved review
+    conditions:
+      - "#approved-reviews-by>=1"
+      - status-success=Travis CI - Pull Request
+      - base=master
+      - label!=work-in-progress
+      - label=ready-to-merge
+    actions:
+      merge:
+        method: rebase
+        rebase_fallback: squash


MVrachev:
What if there are multiple commits into the pull request then what message will be merged? 
The message of the first commit? All messages from all commit related to the pr?

joshuagl:
I like the idea of having an explicit label to tag the PR author is ready to have their work merged (and a separate explicit label to indicate something really isn't ready), it's worth noting however the [GitHub help pages](https://help.github.com/en/articles/about-labels) indicate that only those with write access to the repo can apply a label.

joshuagl:
> What if there are multiple commits into the pull request then what message will be merged?
> The message of the first commit? All messages from all commit related to the pr?

The configuration is to use a rebase, thus all of the commits in the PR will be applied on top of the master branch. It would be good practice to only apply the "ready-to-merge" label once the request submitter has squashed their commits as desired.

It would be good to update CONTRIBUTING.md before we land this change.

ericwb:
@MVrachev So yes, this app would merge using rebase first and squash as a backup.  I assume for the squash it preserves all commit messages.

@joshuagl Yep, the ready-to-merge would be used for the maintainers to use to finalize a PR ready to be merged.  So if one maintainer approved, but wanted a second opinion, he/she would have that option.  I definitely think the contributing doc could use a lot of updates.  I don't mind adding something in the doc about Mergify, but I'm not familiar enough how well it will work yet.  So might be better to update the doc in a separate PR

joshuagl:
I'm OK with CONTRIBUTING in a separate PR.

MVrachev:
So, if I understand correctly the pr workflow will be the following:
1) I open pr for a review
2) we review the code
3) you give me approval
4) I tag the pr with "ready-to-merge" label
5) the pr will be automatically merged

MVrachev:
Another question is can we make Mergify to use only the pr message or first commit message for the commit which will be merged?

ericwb:
@MVrachev In your PR workflow, which would be for any user (even maintainers), you wouldn't self mark the label of ready-to-merge.  That is for approvers.  If your PR is not ready, use work-in-progress or PR Drafts.

And to my knowledge, there is no way to configure Mergify to use just the first commit message.  Of course you can also squash your commits yourself using git, which is probably better practice anyway.

joshuagl:
Looks good to me, let's rebase and merge.

PR #181, opened 2019-04-12
Minor documentation tweaks

joshuagl:
A couple of minor docs tweaks laying the groundwork for documenting the TSLint addition

### DIFF
--- a/docs/false_positivies.md
+++ b/docs/false_positivies.md
@@ -15,7 +15,7 @@ Although this line may cause a potential security issue, it will not be reported
 
 
 If you want to read more about annotating the code:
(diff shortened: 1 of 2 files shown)


MVrachev:
Good changes.
It's a good idea if you squash the commits.
Besides that it's a great addtiion.

joshuagl:
Happy to squash the commits into an "Improve documentation formatting" (or similar) if you feel that's better. 
It does seem to me that these are two distinct changes and therefore appropriate as two patches - I wouldn't squash two distinct code changes into one patch and I think the same rules should be applied to all changes to a repo.

MVrachev:
Yes, you are right. I didn't look closely at the changes separately by the commits.

joshuagl:
Force pushed to fix a typo in a commit message.

PR #182, opened 2019-04-15
Update documentation for TSLint

joshuagl:
Refer to TSLint in the README, update the architecture diagram, document
additional initial setup recommendations and include pointers on excluding
JS/TS issues.


docs/false_positivies.md
```diff
@@ -39,3 +39,14 @@ func main(){
...
+## JavaScript/TypeScript Example
+
+```JavaScript
+
+const foo = "alert('hello')";
+eval(foo); // tslint:disable-line
+
+```
+
+See the TSLint documentation for more about rule flags: [TSLint rule flags](https://palantir.github.io/tslint/usage/rule-flags/) and the tslint-config-security documentation to see a list of rules as used by Precaution [tslint-config-security Rules](https://github.com/webschik/tslint-config-security#rules)
```
MVrachev: If you are adding a link to the rules found by the tslint-config-security project then you should add links maybe in a separate pr with links for the rules of Bandit and Gosec. We can do that so we will be consistent.

joshuagl: The main reason I thought it was relevant here is because the security rules are provided by a separate project to the main linter. If I click through to Bandit or Gosec's documentation I can browse around and find the security rules, if I click through to TSLint's documentation I can't.

You're right though, it's good to be consistent.

MVrachev: I think the best solution will be to have a separate doc page called something like:
"Issues found by Precaution" or "Rules used by Precaution" or something like that where we will provide links to the rules of all linters.
What rules we find is an important question and we can be more expressive about it.

If we want in future we can group the rules there. Like we can have 
- SQL/Command injection section
- Weak cryptography section
and etc.

joshuagl: Some high-level guidance and links for rules would be good. I'm wary of trying to document our dependencies too heavily though, that could become a maintenance nightmare - we should mostly link to the upstream documentation.

Did you want me to remove the tslint-config-security-rules link from this PR? What action should I take here?

MVrachev: Maybe would be better if we focus on false-positives in this document and we can create a new document especially for a high-level view of the supported rules and links to them.

joshuagl: OK, sure. I can remove the link to the the tslint-config-security rules and update this PR.

joshuagl: Done as a follow on patch to be squashed once the changes are approved.

joshuagl:
I've squashed the commits, I think this PR is ready-to-merge?

MVrachev:
Yes, it's good.

PR #183, opened 2019-04-15
Add js and ts files into unit tests

MVrachev:
Connected with issue: https://github.com/vmware/precaution/issues/160

When we check if we download the right files it's a good idea to add js
and ts files there too. Same applies in the merge report unit tests.

### DIFF
--- a/test/index.test.js
+++ b/test/index.test.js
@@ -42,6 +42,9 @@ describe('Precaution workflow', () => {
     mockFiles['networking_binding.go'] = fs.readFileSync('test/fixtures/go/src/vulnerable_package/networking_binding.go', 'utf8')
     mockFiles['bad_test_file.go'] = fs.readFileSync('test/fixtures/go/src/vulnerable_package/bad_test_file.go', 'utf8')
     mockFiles['hello.go'] = fs.readFileSync('test/fixtures/go/src/safe/hello_world.go', 'utf8')
+    // Mock the files as empty strings because we don't analyze their content anyway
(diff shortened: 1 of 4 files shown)


MVrachev:
> Generally a commit message should explain what we did and why, how should be explicit in the change itself and where it isn't we can use code comments to help explain. With that in mind I think:
> 
> > I mock the "vulnerable.js" and "vulnerable.ts" files as empty strings because we don't analyze them or use their content in any way.
> 
> is redundant in the commit message and makes more sense as a code comment.

Yes, you are right. I will add it as a comment.

test/index.test.js
```diff
@@ -42,6 +42,10 @@ describe('Precaution workflow', () => {
     mockFiles['networking_binding.go'] = fs.readFileSync('test/fixtures/go/src/vulnerable_package/networking_binding.go', 'utf8')
     mockFiles['bad_test_file.go'] = fs.readFileSync('test/fixtures/go/src/vulnerable_package/bad_test_file.go', 'utf8')
     mockFiles['hello.go'] = fs.readFileSync('test/fixtures/go/src/safe/hello_world.go', 'utf8')
+    // I mock the "vulnerable.js" and "vulnerable.ts" files as empty strings
```
joshuagl: No need to say "I", perhaps:
```// Mock the files as empty strings, we don't analyse their content anyway```

MVrachev: Ohh... I don't know why I do that mistake so many times.
I will fix it.

PR #184, opened 2019-04-15
Move "reference links" title to config.js

MVrachev:
It's better to have that "reference links" title separated from the
Bandit "more info" implementation. That way when we have reference links
from the other linters it will be possible and easier to add them.


joshuagl:
It would be good to update the commit message to say _why_ this is better

MVrachev:
> It would be good to update the commit message to say _why_ this is better

Yes, I updated the commit and pull request messages.

PR #185, opened 2019-04-15
 Add support for TSLint reference links

MVrachev:
Add support for TSLint reference links

Because TSLint doesn't provide us with reference links and doesn't have
an API for such metadata as explained by the developer of
tslint-config-security: webschik/tslint-config-security#17

All rules are documented as part of the README for
tslint-config-security on GitHub, therefore we can use this
documentation to provide reference links.

Another thing we should check is if every rule has a documentation on
the project webpage and do we point to that documentation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,8 +39,10 @@
     "lint": "standard --fix"
   },
   "dependencies": {
+    "cheerio": "^1.0.0-rc.3",
     "fs-extra": "^7.0.0",
     "probot": "^7.4.0",
+    "request": "^2.88.0",
     "tslint": "^5.14.0",
     "typescript": "^3.3.4000",
     "tslint-config-security": "^1.15.0"
--- a/test/external.dependencies.test.js
+++ b/test/external.dependencies.test.js
@@ -1,0 +1,34 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: BSD-2-Clause
+
+const { config } = require('../config')
+const { jsRules } = require('../tslint/tslint.json')
+
+const request = require('request')
+const cheerio = require('cheerio')
+
+describe('Checks for documentation websites', () => {
+  test('Check if TSLint plugin documentation website is online', async () => {
+    request(config.tslintProjectWebsite, async (err, response, _) => {
+      if (err) {
+        throw new Error(err)
+      }
+      expect(response.statusCode === 200 || response.statusCode === 201).toBeTruthy()
+    })
+  })
+
+  test('Check documentation for every rule on the TSLint plugin webpage', async () => {
+    const rules = Object.keys(jsRules)
+    request(config.tslintProjectWebsite, (err, _, body) => {
+      if (err) {
+        throw new Error(err)
+      }
+      let $ = cheerio.load(body)
+      let result = ''
+      for (let ruleName of rules) {
+        result = $('#user-content-' + ruleName).attr('href')
+        expect(result).toEqual('#' + ruleName)
+      }
+    })
+  })
+})
--- a/tslint/tslint_report.js
+++ b/tslint/tslint_report.js
@@ -3,6 +3,24 @@
 
 const { getAnnotation } = require('./tslint_annotations')
 const { countIssueLevels } = require('../annotations_levels')
+const { config } = require('../config')
+
+/**
+ * @param {*} issues the issues found by TSLint
+ */
+function createMoreInfoLinks (issues) {
+  let issuesMap = new Map()
+  let moreInfo = ''
+
+  for (let issue of issues) {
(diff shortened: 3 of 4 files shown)


MVrachev:
> I really like this change. I'm a little worried that one day those entries will be removed from the upstream README... Can we add a test to check that resolving those links actually generates some content (not receiving an error page would probably be enough?)

That's a good point. I will add a unit test for that in this pr.

MVrachev:
I added a unit test checking the status code of the doc website.

test/external.dependencies.test.js
```diff
@@ -0,0 +1,16 @@
...
+const { config } = require('../config')
+const request = require('request')
+
+describe('Checks for documentation websites', () => {
+  test('Check TSLint plugin documentation website', async () => {
+    request(config.tslintProjectWebsite, async (err, response, body) => {
+      if (err) {
+        throw new Error(err)
+      }
+      expect(response.statusCode === 200 || response.statusCode === 201).toBeTruthy()
```
joshuagl: I believe this checks that the GitHub README is present, whereas what I'm more concerned about is whether the sections of the README disappear. Can we try and load one of the reference links and make sure there's content in the DOM where we expect it?

This might be a lot of work and not provide sufficient value to be worth the effort?

MVrachev: Maybe there is a way to do this.
The problem is that if you write you write a unit test for a specific "label" section of the website like this: 

image

the test passes.
I have to find a way how to check for the existing of a specific label on the website.

Also, because we have all rule names hardcoded in our configuration options for TSLint (in tslint/tslint.json, it will be possible to update our tslint/tslint.json file with a new rule and then our unit test will be updated too.

joshuagl: I haven't written JS for about 11 years, but back then I would have used [`document.getElementById()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementById).

We can see from the HTML of tslint-config-security's README.md that the anchor's all have consistent id's following a pattern based on the target name, for example:
`<a id="user-content-tsr-detect-unsafe-regexp" ...>`
and
`<a id="user-content-tsr-detect-child-process" ...>`

package.json
```diff
@@ -41,9 +41,10 @@
   "dependencies": {
     "fs-extra": "^7.0.0",
     "probot": "^7.4.0",
+    "request": "^2.88.0",
     "tslint": "^5.14.0",
-    "typescript": "^3.3.4000",
-    "tslint-config-security": "^1.15.0"
+    "tslint-config-security": "^1.15.0",
```
ericwb: Is there any reason to reorder these two lines?

MVrachev: It was automatically generated that way :D

test/external.dependencies.test.js
MVrachev: I added tests for that using a really great library for that called [cheerio](https://github.com/cheeriojs/cheerio).

joshuagl:
Typo:

> Having documentation or JavaScript/TypeScript issues will educate the
> people on how to prevent future mistakes

Having documentation _of_ JavaScript/TypeScript issues...

The commit message for the second patch needs tweaking also, includes:

> It's also important to keep an eye to the exact reference
> links we are giving as documentation that's why I am checking
> if every reference link we are going to give as output is valid.

Perhaps:
"Check every reference link we may give as output is valid" ?

MVrachev:
> Typo in the commit message for the first patch:
> 
> > Having documentation or JavaScript/TypeScript issues will educate the
> > people on how to prevent future mistakes
> 
> Having documentation _of_ JavaScript/TypeScript issues...
> 
> The commit message for the second patch needs tweaking also, includes:
> 
> > It's also important to keep an eye to the exact reference
> > links we are giving as documentation that's why I am checking
> > if every reference link we are going to give as output is valid.
> 
> Perhaps:
> "Check every reference link we may give as output is valid" ?

I fixed both of them. 
Do you think it's okay now?

joshuagl:
2nd commit has a typo in the commit message, otherwise this LGTM.

> It's important to keep an eye on the TSLint plugin GitHub age.
age -> page

tslint/tslint_report.js
```diff
@@ -3,6 +3,27 @@
...
+  let moreInfo = ''
+
+  for (let issue of issues) {
+    if (issuesMap.has(issue.ruleName) === false) {
+      issuesMap.set(issue.ruleName)
+      const linkToDoc = config.tslintProjectWebsite + '#' + issue.ruleName
+      moreInfo += `[${issue.ruleName}](${linkToDoc})\n`
+    }
+  }
+  if (issuesMap.size <= 0) {
```
MVrachev: We don't need this check because if issues array is empty then it won't go into to for loop and moreInfo will be an empty string.

PR #196, opened 2019-04-17
Make documentation links relative

joshuagl:
Enable relative links so that when running a local Jekyll instance or
browsing the Markdown documentation files from the GitHub file browser
documentation links don't take us away from the GitHub browser.
The Jekyll jekyll-relative-links plugin is enabled by default on GitHub
pages:
https://help.github.com/en/articles/configuring-jekyll-plugins#default-plugins

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,4 +1,6 @@
-# Copyright 2018 VMware, Inc.
+# Copyright 2019 VMware, Inc.
 # SPDX-License-Identifier: BSD-2-Clause
 
 theme: jekyll-theme-slate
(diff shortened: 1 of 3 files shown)


docs/_config.yml
```diff
@@ -1,4 +1,6 @@
 # Copyright 2018 VMware, Inc.
```
MVrachev: Because you are changing a significant part of that file maybe it's a good idea to update the year of the Copyright.

joshuagl: Good catch, thanks.

PR #202, opened 2019-04-24
Add JS/TS dependencies in manual deployment docs

MVrachev:
We miss a really important step in the manual deployment doc and
it is about installing node and JavaScript/TypeScript dependencies
as a whole.
Without this step, Precaution won't run.


joshuagl:
LGTM

PR #207, opened 2019-04-26
Remove redundant async keyword

MVrachev:
Nodejs is single threaded but uses an event loop
for Asynchronous calls.
It has an async/await mechanism which should be used when we have
external service or when we need to wait for a data
(and we don't have immediate access to it).

Practically that translates to the simple idea that if we are calling
one function that is async or if it returns a Promise
then we need to await the result of that function and thus use our
own async function.

### DIFF
--- a/index.js
+++ b/index.js
@@ -96,13 +96,13 @@ async function processPullRequest (pullRequest, context) {
   const response = await apiHelper.getPRFiles(context, number)
   const filesDownloadedPromise = response
     .map(async filename => {
-      const headRevision = apiHelper.getContents(context, filename, ref, pullRequest.head)
+      const headRevision = await apiHelper.getContents(context, filename, ref, pullRequest.head)
 
       // TODO: merge this code with linter-specific path resolution
       if (filename.endsWith('.go')) {
-        cache.saveFile(repoID, id, filename, (await headRevision).data, 'go')
+        cache.saveFile(repoID, id, filename, headRevision.data, 'go')
       } else {
-        cache.saveFile(repoID, id, filename, (await headRevision).data)
+        cache.saveFile(repoID, id, filename, headRevision.data)
       }
 
       return filename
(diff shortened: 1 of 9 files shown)


MVrachev:
I am sorry that I added the async keyword on soo many instances here: https://github.com/vmware/precaution/commit/ce8e0dca10e18b69640d3852f071b7cc250ece68
and now I had to remove them.
At least now we have better knowledge about when and how to use the async/await keywords.

index.js
```diff
@@ -96,13 +96,13 @@ async function processPullRequest (pullRequest, context) {
   const response = await apiHelper.getPRFiles(context, number)
   const filesDownloadedPromise = response
     .map(async filename => {
-      const headRevision = apiHelper.getContents(context, filename, ref, pullRequest.head)
+      const headRevision = await apiHelper.getContents(context, filename, ref, pullRequest.head)
```
joshuagl: It doesn't look like `getContents()` is an `async` function, does it return a promise?

MVrachev: Yes, it returns a promise.
All "GET" calls to GitHub needs await. The reason is that we can have a slow network connection/server.

When we use "POST" call to GitHub initially when we create a check_run we don't need the result immediately and that's why we await later (on line 70).

joshuagl: The requirement for `await` makes sense, I just couldn't see any syntax indicating that `getContents()` returns a promise (the function isn't annotated with the `async` keyword and the octokit/rest docs don't state that it returns a Promise either) - as I'm a JS novice I wanted to check.

MVrachev: Yep, better ask because there can be a problem there next time.
When I look at the documentation from VS code I see this:
image
so, at least from my side, it looks well documented.

joshuagl: Nice. I should try and find where VSCode is pulling those docs from.

Issue #190, opened 2019-04-16
Using npm and yarn at the same time


MVrachev:
**Describe the bug**
It doesn't make sense to use two package managers at the same time. This can lead to many complications in the future and it's just unnecessary complication.

We should investigate the pluses of npm and yarn.

**Expected behavior**
Use of only one package.

PR #208, opened 2019-04-26
Remove redundant second package manager - yarn

MVrachev:
From the start of the development of Precaution we are
using two package managers - yarn and npm.
This is meaningless given that one of them is enough.
This situation brings both complexities and sometimes
build failures and other problems.

Examples when we update package.json but not yarn.lock
and that resolves in build failures:
- https://github.com/vmware/precaution/commit/6a702fdc26df9228ac9e306d9af3b07872faf2c3
- https://github.com/vmware/precaution/commit/8550ef2d9aae2a558797aaba6dcde6442c91b6ba

It will be better to use npm because npm is distributed with Node.js -
which means that when you download Node.js,
you automatically get npm installed on your computer.

Another reason is that most of the dependencies we are adding are npm
modules and thus it makes sense to use npm to handle them.

Closes: https://github.com/vmware/precaution/issues/190

PS: I tried to remove all instances of yarn. 
If you see any other please let me know.


MVrachev:
> LGTM. Nit: Start putting Closes: # at the bottom of the commit message, not the top.

Done.

Issue #179, opened 2019-04-12
Config options to exclude/include files


MVrachev:
**Is your feature request related to a problem? Please describe.**
It's important that our users will be able to exclude test or dependencies folders.
We can provide them with a config file and options for that.

**Describe the solution you'd like**
Maybe we can use precaution.json for that since we are using JavaScript and it will be easier to serialize it.

**Describe alternatives you've considered**
Maybe we can use a different file like yml file.

MVrachev:
As @ericwb said it will be better to use YAML config file instead of JSON config file because it's more human readable even thought JSON will be easier to parse than YAML.

PR #213, opened 2019-04-30
Add configuration file support and exclude option

MVrachev:
Our users need a way to configure Precaution for their needs.

The first important option to enable will be the "exclude" option.
That way the users can exclude their tests folders and files.

We decided to give globs as a way for the users to be more flexible
when they are creating exclude rules.

Because YAML allows you to use comments and it's more human readable
than JSON we decided to use .precaution.yaml as a config file.

Closes: https://github.com/vmware/precaution/issues/179

### DIFF
--- a/config.js
+++ b/config.js
@@ -12,7 +12,8 @@ const config = {
   moreInfoTitle: 'For more information about the issues follow the links: \n',
   tslintProjectWebsite: 'https://github.com/webschik/tslint-config-security',
   numFilesPerPage: 30,
-  numAnnotationsPerUpdate: 50
+  numAnnotationsPerUpdate: 50,
+  configFilePath: '.precaution.yaml'
 }
 
 const annotationsLevels = {
--- a/filter.js
+++ b/filter.js
@@ -1,0 +1,47 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: BSD-2-Clause
+
+const { config } = require('./config')
+
+const minimatch = require('minimatch')
+const path = require('path')
+
+/**
+ * Checks whether a file should be excluded from processing
+ * @param {String} filePath path of file to check
+ * @param {String[]} excludes array of exclusion rules to compare to
+ */
+function isExcluded (filePath, excludes) {
+  let ok = false
+  for (let excludeRule of excludes) {
+    excludeRule = path.normalize(excludeRule)
+    if (minimatch(filePath, excludeRule, { matchBase: true })) {
+      ok = true
+      break
+    }
+  }
+  return ok
+}
+
+/**
+ * Filters the listed files and returns only the files relevant to us
+ * @param {*} rawData the raw output from list files API call
+ * @param {String[]} excludes excludes globs provided by the user to filter files/folders
+ * @return {Promise[]<any>} the filtered GitHub response
+ */
+function filterData (rawData, excludes) {
+  return rawData.data
+    .filter(file =>
+      config.fileExtensions.reduce(
+        (acc, ext) => acc || file.filename.endsWith(ext),
+        false
+      )
+    )
+    .filter(fileJSON => fileJSON.status !== 'removed')
+    .map(fileInfo => {
+      return fileInfo.filename
+    })
+    .filter(filePath => !isExcluded(filePath, excludes))
+}
+
+module.exports.filterData = filterData
--- a/filter.js
+++ b/filter.js
@@ -26,7 +26,7 @@ function isExcluded (filePath, excludes) {
 /**
  * Filters the listed files and returns only the files relevant to us
  * @param {*} rawData the raw output from list files API call
- * @param {String[]} excludes excludes globs provided by the user to filter files/folders
+ * @param {String[]} excludes array of exclusion rules in glob format provided by the user
  * @return {Promise[]<any>} the filtered GitHub response
  */
 function filterData (rawData, excludes) {
@@ -45,3 +45,4 @@ function filterData (rawData, excludes) {
 }
 
 module.exports.filterData = filterData
+module.exports.isExcluded = isExcluded
--- a/github_api_helper.js
+++ b/github_api_helper.js
@@ -31,10 +31,11 @@ function inProgressAPIresponse (context, headSha) {
  * Get Precaution config file contents as raw data,
  * if that file exists.
  * @param {import('probot').Context} context Probot context
+ * @param {String} customConfigPath Optional (default value config.configFilePath)
  * @return {Promise<any>} GitHub response
  * See https://developer.github.com/v3/repos/contents/#get-contents
  */
-async function getConfigFile (context) {
+async function getConfigFile (context, customConfigPath = config.configFilePath) {
   let { owner, repo } = context.repo()
   // GitHub doesn't provide "checkIfFileExists" API endpoint.
   // That's why we should try to download the file and if we have
@@ -43,7 +44,7 @@ async function getConfigFile (context) {
     return await context.github.repos.getContents({
       owner,
       repo,
-      path: config.configFilePath,
+      path: customConfigPath,
       headers: { accept: rawMediaType }
     })
   } catch (error) {
--- a/github_api_helper.js
+++ b/github_api_helper.js
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 
 const { config } = require('./config')
+const { filterData } = require('./filter')
 
 const rawMediaType = 'application/vnd.github.v3.raw'
 
@@ -27,24 +28,41 @@ function inProgressAPIresponse (context, headSha) {
 }
 
 /**
- * Filters the listed files and returns only the files relevant to us
- * @param {*} rawData the raw output from list files API call
- * @return {Promise[]<any>} the filtered GitHub response
+ * Get Precaution config file contents as raw data,
+ * if that file exists.
+ * @param {import('probot').Context} context Probot context
+ * @return {Promise<any>} GitHub response
+ * See https://developer.github.com/v3/repos/contents/#get-contents
  */
-function filterData (rawData) {
-  return rawData.data.filter(file => config.fileExtensions.reduce((acc, ext) => acc || file.filename.endsWith(ext), false))
-    .filter(fileJSON => fileJSON.status !== 'removed').map(fileInfo => {
-      return fileInfo.filename
+async function getConfigFile (context) {
+  let { owner, repo } = context.repo()
+  // GitHub doesn't provide "checkIfFileExists" API endpoint.
+  // That's why we should try to download the file and if we have
+  // 404 error code the file doesn't exist in the repository.
+  try {
+    return await context.github.repos.getContents({
+      owner,
+      repo,
+      path: config.configFilePath,
+      headers: { accept: rawMediaType }
     })
+  } catch (error) {
+    if (error.code === 404) {
+      return null
+    } else {
+      throw new Error(error)
+    }
+  }
 }
 
 /**
  * Get list of files modified by a pull request
  * @param {import('probot').Context} context Probot context
  * @param {number} number the pull request number inside the repository
+ * @param {Object} configObj configuration object populated by a config file
  * @returns {String[]} paths to the diff files from the pull request relevant to us
  */
-async function getPRFiles (context, number) {
+async function getPRFiles (context, number, configObj) {
   const { owner, repo } = context.repo()
 
   // See https://developer.github.com/v3/pulls/#list-pull-requests-files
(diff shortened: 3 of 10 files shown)


index.js
```diff
@@ -56,12 +56,16 @@ async function runLinterFromPRData (pullRequests, context, headSha) {
   const checkRunResponse = apiHelper.inProgressAPIresponse(context, headSha)
 
   try {
+    // For now only deal with one PR
+    const PR = pullRequests[0]
+
+    // Loads the Precaution configuration file
+    const rawObj = await apiHelper.getConfigFile(context, cache.getBranchPath(repoID, PR.id))
+    const configObj = rawObj ? JSON.parse(rawObj.data) : []
```
joshuagl: Here we're assigning a JavaScript object `{}` if rawObj is non-null or an array `[]` otherwise. We should be consistent and assign the same type to this variable in both instances.

github_api_helper.js
```diff
@@ -27,24 +28,41 @@ function inProgressAPIresponse (context, headSha) {
...
+  try {
+    return await context.github.repos.getContents({
+      owner,
+      repo,
+      path: config.configFilePath,
+      headers: { accept: rawMediaType }
     })
+  } catch (error) {
+    if (error.code === 404) {
+      return undefined
```
joshuagl: it would be cleaner if we always returned the same type from this function, let's return an empty array `[]` rather than `undefined`. Is there some JS convention I'm unaware of that makes returning `undefined` better here?

filter.js
```diff
@@ -0,0 +1,54 @@
...
+const minimatch = require('minimatch')
+const path = require('path')
+
+/**
+ * Checks, whenever a file should be excluded by the,
+ * excludes provided by the user
+ * @param {String} filePath path to the file which will be checked
+ * @param {String[]} excludes array of exclude rules against whom
+ * the file will be checked
+ */
```
joshuagl: I'd suggest we tighten the language up a bit here:
```JavaScript
/**
 * Checks whether a file should be excluded from processing
 * @param {String} filePath path of file to check
 * @param {String[]} excludes array of exclusion rules to compare to
 */
```

filter.js
```diff
@@ -0,0 +1,54 @@
...
+        (acc, ext) => acc || file.filename.endsWith(ext),
+        false
+      )
+    )
+    .filter(fileJSON => fileJSON.status !== 'removed')
+    .map(fileInfo => {
+      return fileInfo.filename
+    })
+    .filter(filePath => {
+      if (excludes.length > 0) {
```
joshuagl: I believe this check is redundant? Won't the `for` loop in `isExcluded` simply skip straight to returning the default, false, value for ok if we pass an empty array for excludes?

MVrachev: Yep, I checked and tested and you are right.

github_api_helper.js
MVrachev: First, it's important to notice that because this function is async it will always return a promise:

_The word “async” before a function means one simple thing: a function always returns a promise. Even If a function actually returns a non-promise value, prepending the function definition with the “async” keyword directs JavaScript to automatically wrap that value in a resolved promise._
from https://javascript.info/async-await

So, even if I return '[]' in that situation we would return basically Promise<[]> instead of Promise<undefine> where in <> is the value of the promise. 
I used "undefined" here because in JavaScript when a variable doesn't have any value is marked as "undefined" and because if there is no config file in the repo then confObj makes sense to be undefined.

Another bonus is that it's easy to check if a value has undefined value like here in index.js
```
// Loads the Precaution configuration file
const rawObj = await apiHelper.getConfigFile(context, cache.getBranchPath(repoID, PR.id))
const configObj = rawObj ? JSON.parse(rawObj.data) : []
```

index.js
MVrachev: With this, we are checking if the rawObj is null, undefined or blank:
https://stackoverflow.com/questions/5515310/is-there-a-standard-function-to-check-for-null-undefined-or-blank-variables-in

filter.js
MVrachev: A good suggestion. It's hard for me to write short descriptions for the parameters and return values.
I will try better. :))

github_api_helper.js
joshuagl: Ah, OK. This is me not being JS savvy then. I had assumed `[] ? 'foo' : 'bar'` would give 'bar' much as `undefined ? 'foo' : 'bar'` does, much as in Python an empty array evaluates to `False`, but reading your comment and checking in a JS console I realise I assumed incorrectly.

index.js
joshuagl: Based on the discussion above on `undefined` vs. `[]` I agree that `undefined` (or `null`?) seems to make sense here

MVrachev: I read through some opioons what is `null` and `undefined`:
image
https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript

and I think in this situation null is more appropriate because we are initializing the config obect with value of null which not `undefined` anymore because it's defined as null. :D

config.js
```diff
@@ -12,7 +12,8 @@ const config = {
   moreInfoTitle: 'For more information about the issues follow the links: \n',
   tslintProjectWebsite: 'https://github.com/webschik/tslint-config-security',
   numFilesPerPage: 30,
-  numAnnotationsPerUpdate: 50
+  numAnnotationsPerUpdate: 50,
+  configFilePath: '.precaution.json'
```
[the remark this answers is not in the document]
MVrachev: Yes, but JavaScript standard doesn't allow us this.
The build of Travis CI will fail because of that.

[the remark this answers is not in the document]
MVrachev: Yes, but the JavaScript standard doesn't allow us this.
The build of Travis CI will fail because of that.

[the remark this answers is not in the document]
ericwb: I do it all the time in javascript.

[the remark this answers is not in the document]
ericwb: Yes, it's allowed.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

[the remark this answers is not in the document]
MVrachev: I mean [JavaScript standard](https://standardjs.com/) guidelines we are following.
I get a style error
image

and when I run `npm test` I get this: 
image

which means that the build will fail because of that.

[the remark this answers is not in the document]
ericwb: Ah, I see.  Well, I recommend we remove that check from eslint.  Comma dangles are very useful for source control diffs.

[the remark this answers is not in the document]
ericwb: Oh this is standardjs, which doesn't allow config.

joshuagl:
Trying to prompt Travis with a close/reopen

joshuagl:
Go Travis, go!

filter.js
```diff
@@ -0,0 +1,47 @@
...
+      break
+    }
+  }
+  return ok
+}
+
+/**
+ * Filters the listed files and returns only the files relevant to us
+ * @param {*} rawData the raw output from list files API call
+ * @param {String[]} excludes excludes globs provided by the user to filter files/folders
```
joshuagl: Maybe something like
```JS 
/**
* @param {String[]} excludes array of exclusion rules in glob format provided by the user
*/ 
```

github_api_helper.js
joshuagl: Based on the other discussion around `undefined` vs `null` we should probably be returning `null` here

test/config.options.test.js
```diff
@@ -0,0 +1,21 @@
...
+    const files = ['tests/func_test.js', 'workflow_test.go', 'src/cmd/main.py']
+    files.forEach(file => {
+      expect(isExcluded(file, confObj.exclude)).toBeTruthy()
+    })
+  })
+
+  test('No excluded files test', () => {
+    const files = ['../tests/', 'workflow.go', 'src/cmd/untilities.py']
+    files.forEach(file => {
+      expect(!isExcluded(file, confObj.exclude)).toBeTruthy()
```
joshuagl: As above, let's use `toBe()` rather than `toBe[Truthy|Falsy]()`, furthermore rather than saying this should not `!` be `true`, let's just say it should be false:
```JS
expect(isExcluded(file, confObj.exclude)).toBe(false);
```

filter.js
```diff
@@ -45,3 +45,4 @@ function filterData (rawData, excludes) {
 }
 
 module.exports.filterData = filterData
+module.exports.isExcluded = isExcluded
```
joshuagl: We're only exporting this so that the tests can see it to test, right? Is there a way we can export this function _only_ to the tests?

github_api_helper.js
```diff
@@ -31,10 +31,11 @@ function inProgressAPIresponse (context, headSha) {
  * Get Precaution config file contents as raw data,
  * if that file exists.
  * @param {import('probot').Context} context Probot context
+ * @param {String} customConfigPath Optional (default value config.configFilePath)
```
joshuagl: If this parameter will only ever be overridden by our tests I think it's an indicator of a deficiency in our test setup.

I assume this is because we don't actually have a git repo with expected files, we're mocking a repo? Is it worth writing a generic function for our tests that can populate a local git repo with some files, then we can have tests create the repo in setup and delete it in teardown?

Note: this is a genuine question, not me saying we should do this.

config.js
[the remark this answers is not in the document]
ericwb: I tried using standard on another project and immediately ran into this bug that has gone unfixed for 10 months.  So my confidence on using standard is not good as a result.

https://github.com/standard/standard/issues/1165

filter.js
MVrachev: I made small research and I don't see a native way in Node.js to export one function only for one file.
I only saw this library which I am not sure how will help us with this: 
https://github.com/jhnns/rewire

Also, I am not sure is it worth it to add a new dependency only because of this.

github_api_helper.js
MVrachev: Right now we are mocking GitHub simulating their responses and behavior as much as we could. 
We use hardcoded responses given values and files to verify that our mocked GitHub version returns the result we expect.

test/config.options.test.js
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: BSD-2-Clause
+
+const confObj = require('./fixtures/config_files/sample_config.json')
+const { isExcluded } = require('../filter/')
+
+describe('Config options tests', () => {
+  test('Exclude files test', () => {
+    const files = ['tests/func_test.js', 'workflow_test.go', 'src/cmd/main.py']
+    files.forEach(file => {
+      expect(isExcluded(file, confObj.exclude)).toBeTruthy()
```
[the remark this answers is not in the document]
MVrachev: A really good point.

filter.js
joshuagl: Agreed, it's not worth adding a dependency for this.

github_api_helper.js
joshuagl: Understood. My suggestion may be too much work as repo contents aren't really what we care about testing. The main thing I wanted to note is that we're adding function parameters just to support testing and that doesn't quite _feel_ right. Something to keep in mind, perhaps.

config.js
ericwb: I do believe that YAML is a better human friendly config format than JSON.   I understand it's easier for a node.js app to consume JSON files, but YAML is better choices for users.

MVrachev: Yea, maybe you are right. Because YAML gives you comments which are quite useful many times.
Also, it's just more readable and not so strict.

MVrachev: I updated my pr with `.precaution.yaml` instead of `.precaution.json`.

github_api_helper.js
```diff
@@ -31,10 +31,11 @@ function inProgressAPIresponse (context, headSha) {
...
  * @return {Promise<any>} GitHub response
  * See https://developer.github.com/v3/repos/contents/#get-contents
  */
-async function getConfigFile (context) {
+async function getConfigFile (context, customConfigPath = config.configFilePath) {
```
joshuagl: It's strange that we introduce a new function in commit one of this PR then immediately change it in commit two of the same PR. That's a perfectly natural way for code to evolve during development but not necessary to expose in the revision control history.

github_api_helper.js
```diff
@@ -43,12 +44,12 @@ async function getConfigFile (context) {
...
       owner,
       repo,
-      path: config.configFilePath,
+      path: customConfigPath,
       headers: { accept: rawMediaType }
     })
   } catch (error) {
     if (error.code === 404) {
-      return undefined
+      return null
```
joshuagl: OOI we had a discussion thread on the first patch in this PR where you concluded that `undefined` made more sense than `null` - what changed? Would this change be more reasonable as part of that first patch?

joshuagl:
@browne want to take a last look at this PR before it merges?

Browne:
> @Browne want to take a last look at this PR before it merges?

@joshuagl - Definitely tagged the wrong person here 😺

joshuagl:
My mistake, apologies for the inconvenience @Browne

joshuagl:
Giving @ericwb a chance for another look before this merges.

github_api_helper.js
MVrachev: This is a change was introduced when we added our unit tests and that's why I change it in the commit where I added the unit tests.

github_api_helper.js
MVrachev: Yes, that was a mistake. 
I fixed it.

Issue #81, opened 2018-12-21
Respect the linter configs files


MVrachev:
We use different linters for Precaution and I think we should respect the configuration files the people are using for those linters.

Also I think that this will give the users an opportunity to set up a specific for a linter options when using Precaution.

**Describe the solution you'd like**
If there is a config file for a specific linter - respect it and use it instead of the default settings.

ericwb:
I'm not sure we should do this as it will become confusing to users if they use both bandit config and precaution config.

MVrachev:
Yep, maybe the best way to do it will be with single .precaution.ymal file in which the users can give Bandit, Gosec specific options and which linters they want to use for Precaution etc.

PR #218, opened 2019-05-02
Document our decision to ignore tool specific configuration files

joshuagl:
We are opting to ignore tool specific configuration files which may exist in
the repository and only use configuration which has been set in precaution.json

This strategy allows us to expose relevant functionality common to all
supported linters in a consistent fashion. Furthermore by not exposing tool
specific configuration via Precaution we don't have to worry about migration
paths should we choose to change the underlying tools in future.

Closes: #81

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -3,6 +3,17 @@
     SPDX-License-Identifier: BSD-2-Clause
 -->
 
+# Overview
+
+Precaution provides simple, automated code reviews for GitHub projects by
+running security analysis tools on all Pull Requests. Precaution achieves this
+by using a variety of tools under the hood without exposing implementation
+details and complexity of the underlying tools to the user.
+
+Concretely, this means that Precaution ignores any tool specific configuration
+files which may exist in the project's repository and performs scans based only
+on Precaution specific configuration.
+
 # Control Flow
 
 ![architecture](architecture.svg)


MVrachev:
Maybe it won't be a bad idea to update the architecture in this pr given that we are saying that we hide the complexity of the underlying linters?

joshuagl:
> Maybe it won't be a bad idea to update the architecture in this pr given that we are saying that we hide the complexity of the underlying linters?

We're explicitly hiding the complexity of the underlying linters from users, not developers. We should abstract the linters in the diagram as previously discussed, but I don't think it needs to happen in this PR?

MVrachev:
> > Maybe it won't be a bad idea to update the architecture in this pr given that we are saying that we hide the complexity of the underlying linters?
> 
> We're explicitly hiding the complexity of the underlying linters from users, not developers. We should abstract the linters in the diagram as previously discussed, but I don't think it needs to happen in this PR?

Ookay, we can use another pr for that.
I think that this text should be added in a future doc for the configuration options.
Maybe we can keep it both places?

joshuagl:
Agreed, I wanted to document the decision we made and prompt a GitHub Pages build by updating a file in docs. When we document our configuration we should include some user-focused text around this too.

joshuagl:
Rebasing the branch to integrate the latest changes from the master branch dismissed the review.

PR #222, opened 2019-05-07
Add doc for exclude option + globs

MVrachev:
Related to pull request https://github.com/vmware/precaution/pull/213
which has to be merged first.

We need documentation for the exclude option and how a user can user
globs.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,0 +1,86 @@
+<!--
+    Copyright 2019 VMware, Inc.
+    SPDX-License-Identifier: BSD-2-Clause
+-->
+
+# Precaution configuration
+
+Precaution uses a variety of linters under the hood without exposing implementation
+details and complexity of the underlying tools to the user.
+In order to prevent conflict with tool specific configuration,
+Precaution ignores any tool specific configuration files which may exist in the
+gtarget git repository and performs scans based solely on Precaution specific configuration.
+
+Precaution provides you with the option to use `.precaution.yaml` configuration file.
+This file **must** be in the root directory of your project.
+
+All paths in the options should be **absolute**.
+
+This file can be configured with the following options:
+
+| Option | Syntax |
+|---|---|
+| [Exclude](#exclude) | [Globs](#globs) |
+
+## Exclude
+
+This option can be used to tell Precaution to exclude files or directories from processing.
+
+This option accepts [Globs](#globs) syntax for defining patterns of files to exclude.
+
+If you want to ignore `src/cmd/main.go` file you will add this to your configuration file:
+```
+exclude:
+- "src/cmd/main.go"
+```
+
+Another use case is when you are using Golang and your dependencies are in
+the `vendor/` folder. Then you can tell Precaution to skip all files
+within the `vendor/` folder like this:
+```
+exclude:
+- "vendor/**"
+```
+
+It's possible that you wouldn't want Precaution to scan your Golang test files. 
+If the suffix in their names is `_test.go` then you can easily exclude them:
+```
+exclude:
+- "*_test.go"
+```
+
+The exclude option supports multiple
+exclude rules:
+```
+exclude:
+- "*_test.go"
+- "vendor/**"
+- "src/cmd/main.go"
+```
+
+## Globs
+
+| Pattern | Meaning |
+|---|---|
+| `abc/**` | Matches everything inside folder abc, including its child folders/subdirectories. |
+| `?` | Matches any character except for `/` one time. |
+| `*` | Matches any character except for `/` multiple times. |
+| `[abc]` | Matches any characters inside the brackets. |
+
+Here are some examples:
+
+| Example | Meaning |
+|---|---|
+| `test/**` | All files under the test folder. |
+| `foo?.py` | Matches `foo1.py`, `fooA.py`, `fooz.go`, etc. |
+| `*.go` | All files with a .go extension. |
+| `[xyz]`  | Matches`x`, `y`, or `z`. |
+
+
+Other glob syntax features supported are:
+- Brace Expansion
+- Extended glob matching
+
+For more detailed explanations of globs see:
+- https://git-scm.com/docs/gitignore
+- http://www.jedit.org/users-guide/globs.html
--- a/docs/index.md
+++ b/docs/index.md
@@ -29,6 +29,7 @@ Precaution currently supports analysis of python files via Bandit and go files v
 
 - [Initial setup](initial_setup.md)
 - [False positives and how to handle them](false_positivies.md)
+- [Configuration options](configuration.md)
 - [Setting up a manual deployment](manual_deployment.md)
 - [Building this documentation locally](local_docs_build.md)
 - [Debugging with VSCode](local_development.md)


docs/configuration.md
```diff
@@ -0,0 +1,90 @@
...
+{
+    exclude: ["*_test.go", "vendor/**", "src/cmd/main.go"]
+}
+```
+
+## Globs
+
+Some of the Precaution configuration options are using Globs as a syntax.
+
+These are just some of the example patterns we are supporting.
```
joshuagl: There are more patterns we support? How does the user find out what they all are?

docs/configuration.md
```diff
@@ -0,0 +1,90 @@
...
+| `foo?.py` | Matches `foo1.py`, `fooA.py`, `fooz.go`, etc. |
+| `[xyz]`  | Matches`x`, `y`, `z`, etc.
+
+
+Other glob syntax features we are supporting are:
+- Brace Expansion
+- Extended glob matching
+
+Read also:
+- https://git-scm.com/docs/gitignore
```
joshuagl: Ahah, here's where `**` is documented 😃

docs/configuration.md
```diff
@@ -0,0 +1,90 @@
+<!--
+    Copyright 2019 VMware, Inc.
+    SPDX-License-Identifier: BSD-2-Clause
+-->
+
+# Precaution configuration
+
+Precaution provides you with the option to use optional `.precaution.json` file.
+This file **should** be in the root directory of your project.
```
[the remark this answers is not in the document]
MVrachev: Yep, it will better to use must. 
You are right.

docs/configuration.md
```diff
@@ -0,0 +1,90 @@
...
+
+All paths provided in the options are processed as if the
+they are from the root directory.
```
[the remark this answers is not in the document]
MVrachev: Agreed.

docs/configuration.md
```diff
@@ -0,0 +1,90 @@
...
+|---|---|
+| [Exclude](#exclude) | [Globs](#globs) |
+
+## Exclude
+
+This option can be used to tell Precaution to exclude files or folders from processing.
+
+If you want to exclude a specific file from the Precaution scan you can provide
+the full path to it starting from the root directory of your project.
+Then if you want to ignore `src/cmd/main.go` file you will have this `.precaution.json` file:
```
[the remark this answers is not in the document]
MVrachev: Hmm...no it's not the same.
I test that and when you have a path starting with a slash it's not the same like the path without the starting slash.
I use `path.normalize()` function to normalize the given path. it removes the extra `.` and `..`
https://nodejs.org/api/path.html#path_path_normalize_path
But it looks like it doesn't remove the starting slash in the path. 
Do you think we should especially handle that situation?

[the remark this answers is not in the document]
joshuagl: That makes sense. I don't think it's necessary to remove an initial `/`, instead we should clarify the documentation. 

I'd probably try and remove the 'root' term completely, as it is a bit ambiguous in this context.
Perhaps something like:
> To exclude a file from the Precaution scan provide the relative path from the base directory of your project.

docs/configuration.md
```diff
@@ -0,0 +1,90 @@
...
+    exclude: ["src/cmd/main.go"]
+}
+```
+
+For example, if you are using Golang and your dependencies are in
+a `vendor/` folder then you can tell Precaution to skip all files
+within the `vendor/` folder with this `.precaution.json` file:
+```
+{
+    exclude: ["vendor/**"]
```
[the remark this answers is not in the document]
MVrachev: Yes, it is documented in the .gitignore documentation.
Also, the library we are using to parse the Globs - minimatch has gitignore listed as reference link: 
image

https://github.com/isaacs/minimatch

docs/configuration.md
```diff
@@ -0,0 +1,90 @@
...
+## Globs
+
+Some of the Precaution configuration options are using Globs as a syntax.
+
+These are just some of the example patterns we are supporting.
+
+| Pattern | Meaning
+|---|---|
+| `*` | Matches everything except slashes. |
+| `abc/**` | Matches everything inside folder abc |
```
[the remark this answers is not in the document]
MVrachev: Yes, I just tested it.

docs/configuration.md
MVrachev: Yep, I saw it from gitignore and from minimatch docs there they mention it a few times.

docs/configuration.md
MVrachev: I mentioned:
- Brace Expansion
- Extended glob matching
below too because in the minimatch docs they say those are supported too.

docs/configuration.md
```diff
@@ -0,0 +1,96 @@
...
+# Precaution configuration
+
+Precaution is using a variety of tools under the hood without exposing
+implementation details and complexity of the underlying tools to the user.
+
+Concretely, this means that Precaution ignores any tool specific configuration
+files which may exist in the project's repository and performs scans based only
+on Precaution specific configuration.
+
+Precaution provides you with the option to use optional `.precaution.json` file.
```
ericwb: Why JSON for config and not YAML?  YAML is much more human friendly readable/writable config nowadays.

docs/configuration.md
```diff
@@ -0,0 +1,96 @@
...
+    SPDX-License-Identifier: BSD-2-Clause
+-->
+
+# Precaution configuration
+
+Precaution is using a variety of tools under the hood without exposing
+implementation details and complexity of the underlying tools to the user.
+
+Concretely, this means that Precaution ignores any tool specific configuration
+files which may exist in the project's repository and performs scans based only
```
ericwb:
```suggestion
files which may exist in the project's repository and performs scans based solely
```

docs/configuration.md
```diff
@@ -0,0 +1,96 @@
...
+on Precaution specific configuration.
+
+Precaution provides you with the option to use optional `.precaution.json` file.
+This file **must** be in the root directory of your project.
+
+All paths in the options are processed as if they are relative to the root directory.
```
ericwb: Isn't this the same as saying the paths should be absolute rather than relative?  Using the word relative to explain an absolute path might trip up some folks.

docs/configuration.md
```diff
@@ -0,0 +1,96 @@
...
+{
+    exclude: ["vendor/**"]
+}
+```
+
+You can use the same syntax if you are using Python
+and you want to exclude all test files in your `test/` folder.
+
+Another possible scenario is when you want to exclude all Golang test files.
+You can achieve this with this `.precaution.json` file:
```
ericwb: Kinda redundant to mention `.precaution.json` again.  This is the fourth occurrence.

docs/configuration.md
```diff
@@ -0,0 +1,96 @@
...
+```
+{
+    exclude: ["*_test.go"]
+}
+```
+
+The exclude option in the `.precaution.json` file supports multiple exclude rules:
```
ericwb: fifth reference to precaution.json

docs/configuration.md
```diff
@@ -0,0 +1,96 @@
...
+```
+{
+    exclude: ["*_test.go", "vendor/**", "src/cmd/main.go"]
+}
+```
+
+## Globs
+
+Some of the Precaution configuration options are using Globs as a syntax.
```
ericwb: We only have 1 option so far.  And this statement doesn't really add value since we already show a syntax of globs above.

docs/configuration.md
ericwb: Agree, this statement doesn't add value.

docs/configuration.md
```diff
@@ -0,0 +1,96 @@
...
+
+| Example | Meaning
+|---|---|
+| `*.go` | All go files. |
+| `test/**` | All files under the test folder. |
+| `foo?.py` | Matches `foo1.py`, `fooA.py`, `fooz.go`, etc. |
+| `[xyz]`  | Matches`x`, `y`, or `z`. |
+
+
+Other glob syntax features we are supporting are:
```
ericwb: Avoid using pronouns, we and I
```suggestion
Other glob syntax features supported are:
```

docs/configuration.md
```diff
@@ -0,0 +1,96 @@
...
+| Option | Syntax |
+|---|---|
+| [Exclude](#exclude) | [Globs](#globs) |
+
+## Exclude
+
+This option can be used to tell Precaution to exclude files or folders from processing.
+
+If you want to exclude a specific file from the Precaution scan you can provide
+the full path to it starting from the root directory of your project.
```
ericwb: This is the 2nd time making this point.

docs/configuration.md
MVrachev: I agreed with you in the past pull request:
https://github.com/vmware/precaution/pull/213#pullrequestreview-237625980

docs/configuration.md
MVrachev: Yea, maybe you are right. I will change it.

docs/configuration.md
MVrachev: Okay, I will remove it.

docs/configuration.md
MVrachev: Maybe it will be better to remove that sentence as a whole. You are right - it doesn't bring any value.

docs/configuration.md
```diff
@@ -0,0 +1,85 @@
...
+The exclude option in the `.precaution.yaml` file supports multiple
+exclude rules:
+```
+exclude:
+- "*_test.go"
+- "vendor/**"
+- "src/cmd/main.go"
+```
+
+## Globs
```
joshuagl: It feels a bit awkward going straight from a discussion of the exclude option to some documentation of globs without tying the two together. Perhaps it would make more sense to have the globs section as a sub-section of the Exclude section (at least whilst it's the only option)?

```suggestion
### Globs


Rules specified in the exclude option support globbing to match patterns of files, rather than having to hard-code a list. The following table defines pattern formats supported by Precaution configuration files:
```

docs/configuration.md
```diff
@@ -0,0 +1,85 @@
...
+| `test/**` | All files under the test folder. |
+| `foo?.py` | Matches `foo1.py`, `fooA.py`, `fooz.go`, etc. |
+| `[xyz]`  | Matches`x`, `y`, or `z`. |
+
+
+Other glob syntax features supported are:
+- Brace Expansion
+- Extended glob matching
+
+Read also:
```
joshuagl:
```suggestion
For more detailed explanations of globs see:
```

docs/configuration.md
MVrachev: I thought that we will reference the `Globs` section from this table:
image
which makes it not a context of the `exclude` options section.

docs/configuration.md
MVrachev: I will add it to the first commit because this will make it as a separate commit.

docs/configuration.md
joshuagl: That's the only part of the document that ties the two together, nowhere else is it made explicit that the Exclude option supports Glob syntax. With the current layout it's easy to overlook, thus I'd like us to be more explicit and tie the two sections together more explicitly somehow. Even if that's just mentioning in the Exclude section that the rules support glob syntax.

MVrachev: I will add a line explaining that in the exclude section itself.
That way if we reference that section it will be easily visible what syntax it uses.

MVrachev:
We should merge first pull request number 213 before this one.
We don't want to have a  doc for a feature we don't currently support.

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
...
+# Precaution configuration
+
+Precaution is using a variety of tools under the hood without exposing
+implementation details and complexity of the underlying tools to the user.
+
+Concretely, this means that Precaution ignores any tool specific configuration
+files which may exist in the project's repository and performs scans based solely
+on Precaution specific configuration.
+
+Precaution provides you with the option to use optional `.precaution.yaml` file.
```
ericwb: "option to use optional" sounds awkward because you're effectively saying twice that something is optional

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
...
+
+This file can be configured with the following options:
+
+| Option | Syntax |
+|---|---|
+| [Exclude](#exclude) | [Globs](#globs) |
+
+## Exclude
+
+This option can be used to tell Precaution to exclude files or folders from processing.
```
ericwb: Sometimes this text mentions folders and sometimes it says directories.  Would be good to be consistent with the terms

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
...
+
+You should use the [Globs](#globs) syntax when defining the exclude option.
+
+If you want to ignore `src/cmd/main.go` file you will add this to your configuration file:
+```
+exclude:
+- "src/cmd/main.go"
+```
+
+Another use-case is when you are using Golang and your dependencies are in
```
ericwb:
```suggestion
Another use case is when you are using Golang and your dependencies are in
```

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
...
+the `vendor/` folder, then you can tell Precaution to skip all files
```
ericwb: This is a long run-on sentence

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
...
+within the `vendor/` folder like this:
+```
+exclude:
+- "vendor/**"
+```
+
+It's possible that you wouldn't want Precaution to scan your Golang test files and if the suffix in their names is `_test.go`
```
ericwb: Run-on sentence

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
...
+then you can easily exclude them:
+```
+exclude:
+- "*_test.go"
+```
+
+The exclude option in the `.precaution.yaml` file supports multiple
```
ericwb: no need to mention precaution.yaml again

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
...
+| `*` | Matches everything except slashes. |
+| `abc/**` | Matches everything inside folder abc, including its child folders/subdirectories. |
+| `?` | Matches any character except for `/` one time. |
+| `[abc]` | Matches any characters inside the brackets. |
+
+Here are some examples:
+
+| Example | Meaning
+|---|---|
+| `*.go` | All go files. |
```
ericwb:
```suggestion
| `*.go` | All files with a  .go extension. |
```

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
...
+- "*_test.go"
+- "vendor/**"
+- "src/cmd/main.go"
+```
+
+## Globs
+
+| Pattern | Meaning
+|---|---|
+| `*` | Matches everything except slashes. |
```
ericwb: Forward slashes and backslashes?

Below you mention explicitly '/', so it's unclear here

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
...
+
+| Option | Syntax |
+|---|---|
+| [Exclude](#exclude) | [Globs](#globs) |
+
+## Exclude
+
+This option can be used to tell Precaution to exclude files or folders from processing.
+
+You should use the [Globs](#globs) syntax when defining the exclude option.
```
ericwb: "should" implies you don't necessarily have to.  So is glob syntax required or not?

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
+<!--
+    Copyright 2019 VMware, Inc.
+    SPDX-License-Identifier: BSD-2-Clause
+-->
+
+# Precaution configuration
+
+Precaution is using a variety of tools under the hood without exposing
```
ericwb:
```suggestion
Precaution uses a variety of tools under the hood without exposing
```

docs/configuration.md
MVrachev: You must but also if you don't use special Glob syntax symbols it will work too.
Like if you want to exclude a specific file you can with `src/main.go` but maybe I can't express this in this sentence.

docs/configuration.md
MVrachev: I change it and I rearranged the place of the `*` symbol to be underneath the `?` symbol.

ericwb:
There is no need to put "which has to be merged first." in PR message bodies.  If there is a dependency, create the dependency using a git rebase.

ericwb:
"We need documentation for the exclude option and how a user can user globs." - This statement is strange.  It sounds as if this is an issue, not a fix.

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
+<!--
+    Copyright 2019 VMware, Inc.
+    SPDX-License-Identifier: BSD-2-Clause
+-->
+
+# Precaution configuration
+
+Precaution uses a variety of tools under the hood without exposing
```
ericwb: It's awkward to use "tools" twice in the sentence. Sounds redundant.

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
...
+implementation details and complexity of the underlying tools to the user.
+
+Concretely, this means that Precaution ignores any tool specific configuration
```
ericwb: Why "concretely"?

MVrachev: Because of this sentence is connected with the statement above.
“Precaution uses a variety of linters under the hood without exposing
implementation details and complexity of the underlying tools to the user **_and because of that_** we ignore linter specific configuration files”.

docs/configuration.md
joshuagl: Perhaps something like:
```suggestion
This option accepts [Globs](#globs] syntax for defining patterns of files to exclude
```

MVrachev:
I updated my pr with the suggestion from @joshuagl about the Glob syntax in the exclude option.

joshuagl:
As the Gif is so large and adding large binaries to a git repo isn't good practice, can we instead upload the video to YouTube or similar and embed it in the README?

docs/configuration.md
joshuagl: I like **and because of that** better than having a separate statement starting with "concretely".

I think even better would be to expand on the reasoning a bit.

> Precaution uses a variety of linters under the hood without exposing implementation details and complexity of the underlying tools to the user. Because of that, and in order to prevent conflict with any configuration files which may exist in the target git repository, Precaution ignores any linter specific configuration files.

MVrachev:
I updated my pr with the last suggestions from @joshuagl https://github.com/vmware/precaution/pull/222#discussion_r310164822.

docs/configuration.md
```diff
@@ -0,0 +1,87 @@
+<!--
+    Copyright 2019 VMware, Inc.
+    SPDX-License-Identifier: BSD-2-Clause
+-->
+
+# Precaution configuration
+
+Precaution uses a variety of linters under the hood without exposing
+implementation details and complexity of the underlying tools to the user. Because of that,
```
ericwb: Run-on sentence beginning with "Because of that,".   Also, grammatically you shouldn't start a sentence with Because.

MVrachev: Thanks for your grammar advice!

PR #223, opened 2019-05-07
Setting up for manual_deployment doc tweaks

MVrachev:
One student wants to contribute to Precaution and after
reading through our doc for manual deployment he gave some
suggestions.

It will be useful if we clarify how you can clone Precaution
and that you should run pip from the root directory of the project.
Because if you have never used Python it's possible
to make a mistake.


MVrachev:
Rebased to be up to date with the master branch.

PR #228, opened 2019-05-09
A split of the initial setup doc

MVrachev:
Our initial setup doc which is showed right after the installation
of Precaution is little overcrowded.

The steps to setup Precaution as branch protection rule make the
doc look complicated when it's not.
The details about the branch protection rule should
be in a separate doc.


docs/branch_protection_rule.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Setting up a Branch Protection Rule
+
+
+In order to allow Precaution to prevent pull requests from merging, do the following:
+
+### 1. Create a pull request
+
+You should create a pull request after the installation of Precaution in order to setup Precaution as a branch protection rule.
+
```
joshuagl: Is the reason we need to have a PR because the protection rule can only be configured once the status check has been observed for the repository?

MVrachev: Even if you had installed an app 1 year ago if you haven't created a pull request Precaution wouldn't be visible in the "Require status checks to pass before merging" option:
image

You will have the same situation when you decide to remove Precaution then until you create a pull request Precaution will be visible "Require status checks to pass before merging" option.

joshuagl: Thanks for explaining. Can we expand the documentation a bit more to explain why the user needs to create a pull request and why it doesn't have to be merged?

MVrachev: I added a paragraph explaining those things.
What do you think?

docs/branch_protection_rule.md
```diff
@@ -0,0 +1,45 @@
...
+
+In order to allow Precaution to prevent pull requests from merging, do the following:
+
+### 1. Create a pull request
+
+If you follow all of the steps below without creating a pull request, unfortunately, 
+you won't be able to add Precaution as a required check. 
+GitHub needs a new pull request which will trigger Precaution and
+that way they will update their list of installed apps on your repository.
```
joshuagl: Maybe something like 

> "GitHub needs a new pull request to trigger Precaution and update the list of status checks available for your repository."

 ?

docs/branch_protection_rule.md
joshuagl: Looks good, minor nits then let's merge it.

docs/branch_protection_rule.md
MVrachev: Yep, sounds better.

docs/branch_protection_rule.md
```diff
@@ -0,0 +1,45 @@
...
+
+You can create a simple pull request with no real or significant changes.
+
+This pull request doesn't have to be merged, because it's sole purpose is to update the GitHub servers.
```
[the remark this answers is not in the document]
MVrachev: Don't you think it makes sense to clarify that they don't have to merge the pr? 
We don't state it anywhere.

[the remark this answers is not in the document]
joshuagl: You're right, we don't. We should indeed be explicit about that.

docs/initial_setup.md
```diff
@@ -9,7 +9,8 @@
 
 For best results, prior to using Precaution, it is recommended to run static code analysis tooling on the code base. That way the current code base does not contain known security issues.
 
-The table below lists the languages supported by Precaution and the static analysis tools used by Precaution in support of the language. You should use the appropriate tool from the list for your initial scans before enabling Precaution.
+You should use the appropriate tool from the list for your initial scans before enabling Precaution.
+The table below lists the languages supported by Precaution and the static analysis tools used by Precaution in support of the language.
```
joshuagl: Not sure why this sentence order was flipped? It feels a bit less natural this way around.

MVrachev: I don't remember why I did it but you are right.

PR #229, opened 2019-05-14
Move linters files into one folder

MVrachev:
It doesn't make sense to use two folders for our linters code -
one for the definition of the linters classes and one for the
report and annotation files.

When we have a separate folder in our root directory for every
linter gets overcrowded in the root directory over time.


ericwb:
Needs a rebase

PR #233, opened 2019-05-15
Change the build badge link of Travis CI

MVrachev:
We are using the badge for travis-ci.com instead
of travis-ci.org which we are using right now.


MVrachev:
I am not sure. I haven't changed the admin settings for that.
Maybe @joshuagl would know.

Issue #209, opened 2019-04-26
Using regex to scan for specific credentials


ericwb:
**Is your feature request related to a problem? Please describe.**
See https://github.com/Yelp/detect-secrets for reference, but it would be useful for Precaution to also warn on secrets, Confidential code headers, etc.  So level of detection is done through GitHub token scanning, but not all. 

Note that also detect-secrets operates on all files (json, txt, etc), and not just languages files. There are definitely cases where API keys are stuffed into shell scripts, json files, yaml, etc.

**Describe the solution you'd like**
A new, slightly different type of scanner.  Not really a linter.  And it would scan all types of file extensions, not just language code, so the download of files would need to include everything.

**Describe alternatives you've considered**
Token scanning provided by GitHub does some things by not all.  I don't think it would find a Confidential header whereas detect-secrets could be configured to do so.  But it would be good to research and do a comparison analysis.

There are also probably many other CLI tools to do similar scans. This was one recommended from a blog article.

**Additional context**

ericwb:
This integration could also add some differentiation as compared to similar competing Apps

ericwb:
Note; detect-secrets overlaps with language specific linters somewhat. It's KeywordDetector plugin will detect variables named password or secrets, same as Bandit.  

And many of the other plugins it has could be integrated into language specific linters, since they would be better at detecting using an AST rather than regex.

However, non-language files (*.py, *.go, *.js, *.ts) would be best detected by detect-secrets.  And there are examples of people putting secret keys in json files, yaml files, etc.

ericwb:
So should detect-secrets only be used on files not scanned by other linters?

joshuagl:
I like the idea of `detect-secrets` only scanning files not scanned by other linters, but I don't think it's as cut and dry as not scanning files scanned by other linters:
1) linters may not support secret detection (i.e. I don't think tslint does)
2) a linters secret detection may not be as complete as `detect-secret`'s , for example I don't think many language security linters do high entropy string detection? 

There's a few approaches we could take here:
1. try to ensure (through contributions, where necessary) that all linters we use have comprehensive secret detection:
  i. we could ramp up to this by having detect-secrets exclude not run against a list of file types for which we have confidence the language linter handles secret detection. I think this list would start out at zero and increase as we work with the upstream linters?
2. have some logic to prevent reporting the same/similar issues for the same line of code. This might mean:
  a. having a model to map similar functionality across linters/scanners which will be run against the same code
[...]

PR #235, opened 2019-05-15
Add report functionality for detect-secrets + tests

MVrachev:
This is the reporting functionality for detect_secrets which will
process the results from the detect_secrets detect_secrets scan
and it will create valid GitHub annotation objects for the issues.

All upcoming pull requests to integrate detect_secrets
into Precaution can be found here:
https://github.com/vmware/precaution/issues/209#issuecomment-492964642

Related to: https://github.com/vmware/precaution/issues/209

### DIFF
--- a/linters/detect_secrets/detect_secrets_annotations.js
+++ b/linters/detect_secrets/detect_secrets_annotations.js
@@ -1,0 +1,32 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: BSD-2-Clause
+
+const pathLib = require('path')
+
+const { getAnnotationLevel } = require('../../annotations_levels')
+
+/**
+ * @param {*} issue an issue from which the annotation will be build
+ * @param {String} filePath file path where the issue is located
+ * @return {Object} returns an annotation object as specified here:
+ * https://developer.github.com/v3/checks/runs/#annotations-object
+ */
+function getAnnotation (issue, filePath) {
+  const path = pathLib.normalize(filePath)
+  const startLine = issue.line_number
+  const endLine = startLine
+  const annotationLevel = getAnnotationLevel('HIGH', 'LOW')
+  const title = issue.type
+  const message = `The issue was found by the ${issue.type} plugin in detect-secrets.`
+
+  return {
+    path,
+    start_line: startLine,
+    end_line: endLine,
+    annotation_level: annotationLevel,
+    title,
+    message
+  }
+}
+
+module.exports.getAnnotation = getAnnotation
--- a/linters/detect_secrets/detect_secrets_report.js
+++ b/linters/detect_secrets/detect_secrets_report.js
@@ -1,0 +1,29 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: BSD-2-Clause
+
+const { getAnnotation } = require('./detect_secrets_annotations')
+const { countIssueLevels } = require('../../annotations_levels')
+
+/**
+ * Process detect-secrets output (generate annotations, count issue levels)
+ * @param {*} results detect-secrets json output
+ * @returns {{{Object[]}, Object, String}} Object[] array of annotation objects
+ * see https://developer.github.com/v3/checks/runs/#annotations-object
+ * Object contains the number of errors, warnings and notices found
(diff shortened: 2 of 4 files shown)


linters/detect_secrets/detect_secrets_annotations.js
```diff
@@ -0,0 +1,30 @@
...
+ * @return {Object} returns an annotation object as specified here:
+ * https://developer.github.com/v3/checks/runs/#annotations-object
+ */
+function getAnnotation (issue, filePath) {
+  const path = filePath
+  const startLine = issue.line_number
+  const endLine = startLine
+  const annotationLevel = getAnnotationLevel('HIGH', 'LOW')
+  const title = issue.type
+  const message = `The issue is found by the ${issue.type} plugin in detect-secrets.`
```
ericwb:
```suggestion
  const message = `The issue was found by the ${issue.type} plugin in detect-secrets.`
```

linters/detect_secrets/detect_secrets_report.js
```diff
@@ -0,0 +1,23 @@
+// Copyright 2019 VMware, Inc.
+// SPDX-License-Identifier: BSD-2-Clause
+
+const { getAnnotation } = require('./detect_secrets_annotations')
+const { countIssueLevels } = require('../../annotations_levels')
+
+/**
+ * Process detect-secrets output (generate annotations, count issue levels)
+ * @param {*} results detect-secrets json output
+ */
```
ericwb: Add information on return via @return

linters/detect_secrets/detect_secrets_annotations.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {*} issue an issue from which the annotation will be build
+ * @param {String} filePath file path where the issue is located
+ * @return {Object} returns an annotation object as specified here:
+ * https://developer.github.com/v3/checks/runs/#annotations-object
+ */
+function getAnnotation (issue, filePath) {
+  const path = filePath
+  const startLine = issue.line_number
+  const endLine = startLine
+  const annotationLevel = getAnnotationLevel('HIGH', 'LOW')
```
ericwb: Should the severity always be high and confidence always low for these results?

MVrachev:
> So this PR covers only partial functionality. So what other PRs are you planning to submit? It would be good to get in the practice of documenting your planned PRs in the Issue.

I am doing that already: https://github.com/vmware/precaution/issues/160 I was just not sure where to put my planned pull requests.
I mean should I put them in the existing issue tracker as a comment: https://github.com/vmware/precaution/issues/209 or should I create another issue where it will be more visible the progress?

> You PR message added design questions, but the PR is not really a good place for that.

Yes, you are right. I just thought about those design questions just before I submitted the pr and I didn't realize that they are better to be placed in the issue tracker.
I will fix that.

linters/detect_secrets/detect_secrets_annotations.js
MVrachev: That's something we should discuss in our issue tracker
https://github.com/vmware/precaution/issues/209
I just put the severity and confidence Bandit uses when reporting hardcoded credentials.

linters/detect_secrets/detect_secrets_report.js
MVrachev: I noticed the bandit_report and gosec_report don't have `returns` documentation.
If you approve the way I document the return value of this function I will add the same style for them.

MVrachev:
I moved the design discussion in our issue tracker: https://github.com/vmware/precaution/issues/209#issuecomment-492957066
and I gave planning how many pull requests we need to fully integrate detect-secrets here: https://github.com/vmware/precaution/issues/209#issuecomment-492964642

linters/detect_secrets/detect_secrets_annotations.js
ericwb: You might want to look at https://github.com/lyft/bandit-high-entropy-string. This is a bandit plugin to do detect-secrets like testing.  As part of the plugin, it defines severity and confidence.

MVrachev:
As I explained here https://github.com/vmware/precaution/issues/209#issuecomment-501589208 we decided that we won't use `detect-secret`.

PR #236, opened 2019-05-20
Fix filename extensions in config.fileExtensions to have preceding '.'

joshuagl:
Some of file extensions were listed in the fileExtensions array with a
preceding '.' and others were not. We must be consistent in using the '.'
prefix to all file extensions in the array, otherwise we may have unexpected
behaviours such as the filterData() function in github_api_helper.js
matching all filenames that end with 'js' or 'ts'.

### DIFF
--- a/config.js
+++ b/config.js
@@ -3,7 +3,7 @@
 
 const config = {
   cleanupAfterRun: true,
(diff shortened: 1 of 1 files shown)


MVrachev:
Good catch!
LGTM

Issue #227, opened 2019-05-09
Use docker image for deploying Precaution


MVrachev:
**Is your feature request related to a problem? Please describe.**
It's really frustrating when you try to setup Golang and our Golang dependencies on Heroku.
Even if we manage to use the Heroku go buildpack again there are difficulties because golang is not on Heroku after finishing of the deployment.
That, combined with the fact that Gosec uses a new Golang parser in version 2.0.0 which requires Golang installed, creates a lot of troubles. 

We need a new solution for how we are going to deploy to Heroku.

**Describe the solution you'd like**
Docker image sounds like a reasonable solution.
This will give us the freedom to deploy everywhere and customize our build for our needs.
Another bonus is that it will really useful for our future contributors.

joshuagl:
Having a container image will likely make a switch to Circle CI easier too #226

PR #238, opened 2019-06-06
Add Dockerfile

MVrachev:
As we discussed before #227
there are many benefits to creating a Docker image which will be run
on Heroku. That way we would be free from the limitations provided
by Heroku and we would be able to use Go modules.
Also, if one day we decide to move Precaution on another
hosting platform it will be really easy.

We won't need the ".profile" file anymore because we would install
Gosec in our Dockerfile as all other dependencies.

Closes: https://github.com/vmware/precaution/issues/227

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,55 @@
+# Copyright 2019 VMware, Inc.
+# SPDX-License-Identifier: BSD-2-Clause
+
+FROM photon:3.0
+
+# Install the needed programming languages and tools
+RUN tdnf update -y && tdnf install -y \
+    go \
+    python3 \
+    python3-pip \
+    python3-setuptools \
+    nodejs \
+    git
+
+# Copy precaution in the root directory
+COPY . precaution
+
+# Setup a work directory for all comamnds below
+WORKDIR ./precaution
+
+# Install all Python dependencies
+RUN pip3 install --upgrade pip && pip3 install -r requirements.txt
+
+# Install all JavaScript and TypeScript dependencies
+RUN npm install
+
+# Download the tar.gz gosec file version 2.0.0
+RUN curl -OL https://github.com/securego/gosec/releases/download/2.0.0/gosec_2.0.0_linux_amd64.tar.gz
+
+# We need sha256sum which is not installed by default in toybox but it's available in util-linux.
+# There is a conflict between toybox and util-linux and that's why we have to remove toybox.
+RUN tdnf remove toybox -y && tdnf install util-linux -y
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+RUN curl -OL https://github.com/securego/gosec/releases/download/2.0.0/gosec_2.0.0_linux_amd64.tar.gz
+
+# We need sha256sum which is not installed by default in toybox but it's available in util-linux.
+# There is a conflict between toybox and util-linux and that's why we have to remove toybox.
+RUN tdnf remove toybox -y && tdnf install util-linux -y
+
+# Validate Gosec tar.gz file checksum
+RUN echo "490c2a0434b2b9cbb2f4c5031eafe228023f1ac41b36dddd757bff9e1de76a2b  ./gosec_2.0.0_linux_amd64.tar.gz" | sha256sum -c -
+
+# Install all nessery tools to unzip the gosec tar.gz file
```
joshuagl: > nessery -> necessary

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+# Cleanup of the already unnecessary installed packages
+RUN tdnf remove -y \
+    python3-pip \
+    python3-setuptools \
+    git \
+    util-linux \
+    tar \
+    gzip
+
+# Start application on port given by Heroku
```
ericwb:
```suggestion
# Start the application on the port given by Heroku
```

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+
+# Install all necessary tools to unzip the gosec tar.gz file
+RUN tdnf install -y \
+    tar \
+    gzip
+
+# Unzip and remove the redundant tar.gz file
+RUN tar -xvzf gosec_2.0.0_linux_amd64.tar.gz -C /usr/bin && rm ./gosec_2.0.0_linux_amd64.tar.gz
+
+# Cleanup of the already unnecessary installed packages
```
ericwb:
```suggestion
# Cleanup of unnecessary installed packages
```

merge_reports.js
```diff
@@ -31,7 +31,10 @@ const mergeSummaries = (summaries) => {
  * for object schema see: https://developer.github.com/v3/checks/runs/#output-object
  */
 module.exports = (reports) => {
-  const annotations = reports.map(r => r.annotations).flat()
+  let annotations = []
```
ericwb: This change is unrelated and should not be part of this PR.

joshuagl: This change was merged as part of PR #247

PR #244, opened 2019-06-18
Change architecture diagram

MVrachev:
With the addition of the configuration options, we need to
add those changes to the diagram because currently, it doesn't
represent the Precaution control flow correctly.


docs/architecture.md
```diff
@@ -29,3 +29,5 @@ External linters are spawned asynchronously, their output is redirected to
 a file that is then read and converted to the format expected by GitHub checks.
 
 The results are then merged and send back to GitHub with the appropriate status.
+
+We created that diagram using the draw.io website.
```
ericwb: Avoid using pronouns like "we", "I", etc. Just state the diagram was created using draw.io

MVrachev: Ohhh… sorry. I know I do that mistake often. I will try to be more precise.

MVrachev:
> So the diagram changes might not be obvious what you're trying to convey. Putting "..." is typically used in math or programming to indicate 0 to n. But in a diagram it looks more like a dotted line indicating some kind of relationship between GoSec and TSLint.

Then what options do we have?
Use a magical box called “spawn linters” without explicitly showing that we start a new thread for every linter?
How can we show that we start different threads for every linter and still make the diagram easy to understand?

joshuagl:
We only have the 3 linters for now. Let's drop the ellipses between Gosec and TSLint then merge this change.

MVrachev:
I updated my pr based on the feedback provided by @joshuagl with comment https://github.com/vmware/precaution/pull/244#issuecomment-526512813

Issue #106, opened 2019-01-16
Animated Gif on README


MVrachev:
**Is your feature request related to a problem? Please describe.**

We can create an animated GIF in our README how we use Precaution.
That will attract more users.

**Describe the solution you'd like**

There is a good blog about that:https://blog.github.com/2018-06-29-GIF-that-keeps-on-GIFing/

PR #245, opened 2019-06-20
Add Gif to the README

MVrachev:
It will be really useful if we add a GIF to the Precaution README.
That way we would show how Precaution works in 17 seconds!

Closes: https://github.com/vmware/precaution/issues/106

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@
 Precaution provides simple, automated code reviews for GitHub projects by running
 code linters with a security focus on all pull requests.
 
+![Precaution Demo](./precaution.gif)
+
 GitHub integration is made through the GitHub app interface and the checks API (beta),
 which allows results to be presented directly as inline annotations instead of
 a pass/fail status report.


joshuagl:
[ImageOptim](https://imageoptim.com/mac) wraps a bunch of image compression tools in a macOS GUI

MVrachev:
I tried to export all images in 720 p, them I used ImageOptim as @joshuagl  suggested and in the end I combined them with gifsky. On Gifsky I set it with quality of 50 %.
The result is blurred Gif which is 30 mb in size.

I don't know how to shrink it more.

joshuagl:
As the Gif is so large and adding large binaries to a git repo isn't good practice, can we instead upload the video to YouTube or similar and embed it in the README?

ericwb:
How about using git-lfs for this file?

https://help.github.com/en/articles/collaboration-with-git-large-file-storage
https://help.github.com/en/articles/configuring-git-large-file-storage

joshuagl:
This does seem like a good use-case for Git LFS, we gain the benefits of a Gif embedded in the README without having to worry about the impact of a large file on our git repository.

